PR #4, opened 2022-02-15
Collateral replaces Stakes

markspanbroek:
Introduces a new Collateral design that replaces the existing Stakes contract.

In preparation for #2, we introduce a new design for locking up collateral. The main benefit
of this design is that it allows for unlocking of the collateral of multiple hosts in an O(1)
operation. This will become important when storage requests and offers are moved on-chain,
and selection of an offer will unlock the collateral of all the hosts whose offer has not been selected.


dryajov (approved):
I think this is good enough to merge at least for now. I don't see anything egregious, but in any case I expect a few more iteration over the entire system, thus I see no reason to hold it up for now.

PR #11, opened 2022-04-21
feat: verify proof using BN254 curve

emizzle:
Add support for verifying proofs using the BN254 curve.


contracts/Proofs.sol
```diff
@@ -121,6 +125,252 @@ contract Proofs {
...
+    PublicKey memory spk) internal returns (bool) {
+
+    // TODO: is this needed in solidity?
+    // require(!isEmpty(tau.signature), "Signature cannot be empty");
+
+    // TODO: $tau.t -- how to do this in solidity?
+    // TODO: what's the second G2 point needed, PublicKey.signkey is a G1
+    // point?? Is using the PublicKey.key sufficient?
+    uint hashedMsg = uint(sha256(abi.encodePacked(tau.t)));
+    require(_verifySignature(tau.signature, spk.key, hashedMsg),
```
markspanbroek, line 322: I don't think that we need tau.signature. The parameters from the [storage request](https://github.com/status-im/dagger-contracts/blob/2bf01da728317fd1f3ae8424785f31b53eb4013f/contracts/Marketplace.sol#L125) should be enough. When requesting storage, the client submits these parameters in a signed transaction. So no need to verify the signature here.

emizzle: Even though the transaction is signed, how do we know that the signature is valid if we don't verify it?

contracts/Proofs.sol
```diff
@@ -10,7 +14,7 @@ contract Proofs {
     uint256 __period,
     uint256 __timeout,
     uint8 __downtime
-  ) {
+  ) public {
```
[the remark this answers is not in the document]
emizzle: I think that was to satisfy a <= 0.6.x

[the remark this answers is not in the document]
emizzle: After moving to 0.8.x, it has been removed.

contracts/Proofs.sol
```diff
@@ -121,6 +125,252 @@ contract Proofs {
...
+    //
+    // This matches FR = distinct BNU256 in nim-bncurve:
+    // type
+    //   BNU256* = array[4, uint64]
+    //
+    // Note: for BLS curves (in nim-blscurve),
+    // type
+    //   blst_fp = array[0..5, uint64]
+    // uint64[4] ls; // this is what the data type SHOULD be
+    uint      ls;  // but using a uint256 has the same number of bytes, but
```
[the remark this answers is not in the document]
emizzle: `npm run format` didn't change `uint` to `uint256`, but I'll add it back explicitly.

contracts/Proofs.sol
```diff
@@ -121,6 +125,252 @@ contract Proofs {
...
+  }
+
+  struct QElement {
+    int64     i;
+    BnScalar  v;
+  }
+  function isEmpty(bytes32[96] memory array) internal pure returns (bool) {
+    for(uint i; i< array.length; i++){
+      if(i > 0) {
+        return true;
```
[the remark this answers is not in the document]
emizzle: Good catch, however it was a leftover and no longer used. Has been removed.

contracts/Proofs.sol
```diff
@@ -1,5 +1,9 @@
 // SPDX-License-Identifier: MIT
-pragma solidity ^0.8.0;
+pragma solidity >=0.6.0 <=0.8.13;
```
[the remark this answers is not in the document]
emizzle: I managed to pull in some underlying deps so we can get on 0.8.x.

contracts/Proofs.sol
markspanbroek: As I understand it, the signature is used to verify that the parameters in Tau (public key, name, etc.) that are presented for verification, are the same ones that were generated by the client. We already know that these parameters are correct, because they are stored in the smart contract after they were submitted by the client in a signed transaction. So tau.signature doesn't need to be generated, stored or verified, if I understand correctly. What do you think @cskiraly?

emizzle: Thanks, Mark. Do we need to check that it's the same signature that's stored in the contract?

markspanbroek: > Do we need to check that it's the same signature that's stored in the contract?

I think that we don't need the signature at all. We can retrieve the Tau parameters from the contract instead of getting them via the verifyProof parameter, which guarantees that they have been submitted by the client.

emizzle: Thanks guys. Removed.

emizzle:
This is waiting on https://github.com/status-im/nim-codex/pull/101 to work on x86 before we can get some values that can be used to test `verifyProof` with.

emizzle:
The Bn254 proof verification has been implemented such that other curves (eg BLS12-381) can be substituted quite easily.  Some `Proof.test` tests are failing as expected while we wait for valid proof values from status-im/nim-codex#101.

markspanbroek:
Closing  because we're moving towards ZK based proofs.

PR #13, opened 2022-07-20
Multiple slots

markspanbroek:
Adds multiple slots per storage contract, each of which can be filled by a different host.

A step towards the design in https://github.com/status-im/codex-research/pull/83.


emizzle:
I think we can close this as it's quite broad and mostly implemented. The remainder to be implemented have been captured in other issues.

PR #14, opened 2022-08-16
[marketplace] Add request state and add/remove slot proving based on state

emizzle:
Once a storage request is created by the client, if all the slots are filled before the expiry lapses, the request state of the contract is set to `RequestState.Started`. The lock expiry (that prevents client funds withdrawal) is also extended.

If the expiry lapses before all slots are filled, the request state is considered cancelled, however the state cannot be automatically updated. At this point, the functions `isCancelled` and `isSlotCancelled` will return true. Also at this point, the client can withdraw its funds from the contract, and in that transaction, the request state of the contract is set to `RequestState.Cancelled`.

Proofs cannot be marked as missing when `isSlotCancelled` is true. This is the point at which a storage request's expiry lapsed before all slots were filled.

### NOTES
1. ~~This PR is based off of https://github.com/status-im/dagger-contracts/pull/15, which should probably be merged first and then this PR can be rebased on top of main afterwards.~~

### DIFF
--- a/contracts/Marketplace.sol
+++ b/contracts/Marketplace.sol
@@ -72,6 +72,7 @@ contract Marketplace is Collateral, Proofs {
     _submitProof(slotId, proof);
 
     slot.host = msg.sender;
+    slot.requestId = requestId;
     context.slotsFilled += 1;
     emit SlotFilled(requestId, slotIndex, slotId);
     if (context.slotsFilled == request.ask.slots) {
@@ -119,18 +120,29 @@ contract Marketplace is Collateral, Proofs {
     emit RequestCancelled(requestId);
   }
 
-  /// @notice Rseturn true if the request state is RequestState.Cancelled or if the request expiry time has elapsed and the request was never started.
+  /// @notice Return true if the request state is RequestState.Cancelled or if the request expiry time has elapsed and the request was never started.
   /// @dev Handles the case when a request may have been cancelled, but the client has not withdrawn its funds yet, and therefore the state has not yet been updated.
   /// @param requestId the id of the request
   /// @return true if request is cancelled
   function isCancelled(bytes32 requestId) public view returns (bool) {
-    RequestContext storage context = requestContexts[requestId];
+    RequestContext memory context = requestContexts[requestId];
     return
       context.state == RequestState.Cancelled ||
       (
         context.state == RequestState.New &&
         block.timestamp > requests[requestId].expiry
       );
+  }
+
+  /// @notice Return true if the request state the slot belongs to is RequestState.Cancelled or if the request expiry time has elapsed and the request was never started.
+  /// @dev Handles the case when a request may have been cancelled, but the client has not withdrawn its funds yet, and therefore the state has not yet been updated.
+  /// @param slotId the id of the slot
+  /// @return true if request is cancelled
+  function isSlotCancelled(bytes32 slotId) public view returns (bool) {
+    Slot memory slot = slots[slotId];
+    require(slot.host != address(0), "Slot empty");
+    require(slot.requestId != 0, "Missing request id");
+    return isCancelled(slot.requestId);
   }
 
   function _host(bytes32 slotId) internal view returns (address) {
@@ -225,6 +237,7 @@ contract Marketplace is Collateral, Proofs {
   struct Slot {
     address host;
     bool hostPaid;
+    bytes32 requestId;
   }
 
   event StorageRequested(bytes32 requestId, Ask ask);
--- a/contracts/Marketplace.sol
+++ b/contracts/Marketplace.sol
@@ -76,6 +76,7 @@ contract Marketplace is Collateral, Proofs {
     emit SlotFilled(requestId, slotIndex, slotId);
     if (context.slotsFilled == request.ask.slots) {
       context.state = RequestState.Started;
+      _extendLockExpiry(requestId, block.timestamp + request.ask.duration);
       emit RequestFulfilled(requestId);
     }
   }
--- a/contracts/Marketplace.sol
+++ b/contracts/Marketplace.sol
@@ -116,6 +116,16 @@ contract Marketplace is Collateral, Proofs {
     emit RequestCancelled(requestId);
   }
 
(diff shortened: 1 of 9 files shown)


contracts/Marketplace.sol
```diff
@@ -92,6 +98,61 @@ contract Marketplace is Collateral, Proofs {
...
+  /// @dev Request must be expired, must be in RequestState.New, and the transaction must originate from the depositer address.
+  /// @param requestId the id of the request
+  function withdrawFunds(bytes32 requestId) public marketplaceInvariant {
+    Request memory request = requests[requestId];
+    require(block.timestamp > request.expiry, "Request not yet timed out");
+    require(request.client == msg.sender, "Invalid client address");
+    RequestContext storage context = requestContexts[requestId];
+    require(context.state == RequestState.New, "Invalid state");
+
+    uint256 amount = _price(request);
```
markspanbroek, line 111: This is ok for now, but will have to be changed once we start paying out hosts for the time they fill a slot. The amount that we paid to hosts will then have to be deducted from the price.

contracts/Marketplace.sol
```diff
@@ -92,6 +98,61 @@ contract Marketplace is Collateral, Proofs {
...
+    funds.sent += amount;
+    funds.balance -= amount;
+    token.transfer(msg.sender, amount);
```
markspanbroek, line 114: I would move the transfer to the very end of the function, to confirm to the [check-effects-interations pattern](https://docs.soliditylang.org/en/v0.8.16/security-considerations.html#use-the-checks-effects-interactions-pattern). 

We should probably also check the return value of the transfer function:

```suggestion
    require(token.transfer(msg.sender, amount), "Payment failed);
```

test/Storage.test.js
```diff
@@ -111,6 +112,34 @@ describe("Storage", function () {
...
+    it("fails to mark proof as missing when cancelled", async function () {
+      await storage.fillSlot(slot.request, slot.index, proof)
+      await advanceTimeTo(request.expiry + 1)
+      let missedPeriod = periodOf(await currentTime())
+      await expect(
+        storage.markProofAsMissing(slotId(slot), missedPeriod)
+      ).to.be.revertedWith("Request was cancelled")
+    })
+  })
+  describe("contract state", function () {
```
markspanbroek, line 125: I wonder whether we shouldn't also change `isProofRequired()`, `willProofBeRequired()`, and `proofEnd()` when request is cancelled.

markspanbroek (commented):
Thanks @emizzle, looks good! I have left a few comments/questions.

contracts/Marketplace.sol
```diff
@@ -67,11 +71,13 @@ contract Marketplace is Collateral, Proofs {
...
-    RequestState storage state = requestState[requestId];
     slot.host = msg.sender;
-    state.slotsFilled += 1;
+    slot.requestId = requestId;
+    context.slotsFilled += 1;
     emit SlotFilled(requestId, slotIndex, slotId);
-    if (state.slotsFilled == request.ask.slots) {
+    if (context.slotsFilled == request.ask.slots) {
+      context.state = RequestState.Started;
+      _extendLockExpiry(requestId, block.timestamp + request.ask.duration);
```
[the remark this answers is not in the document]
emizzle: When a client submits a storage request, they ask for files to be stored for a `duration`. That means that proofs must be provided until the end of the contract. If we only require each host to provide proofs for a total of `duration` time starting from the time they fill the slot, then unless they filled the last slot (and thus started the contract), they will not be providing proofs until the end of the contract.

I must admit, providing regular proofs for filled slots BEFORE a contract is started seems odd to me. Perhaps removing this requirement would be a potential solution? This means that hosts will only be required to provide regular proofs once the contract has started (in addition to the proof required and submitted to fill the slot).

test/Storage.test.js
emizzle: I believe this is covered in https://github.com/status-im/dagger-contracts/pull/19.

contracts/Marketplace.sol
emizzle: Added a TODO

test/Storage.test.js
emizzle: `proofEnd` is not covered by state checks for cancelled or finished. `proofEnd` returns a `uint256`, so what should the value be if a contract is cancelled?

contracts/Marketplace.sol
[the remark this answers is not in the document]
markspanbroek: What I mean is that the end of the contract should be `time of request submission + duration`. This means that there are no surprises about the end time of the contract. It is already know when the request is submitted, and does not change based on when the slots were filled.

> What was the reasoning behind providing regular proofs for filled slots BEFORE a contract is started?

Hosts make costs from the moment that they fill a slot, and they're compensated for that because they get paid for every second that they fill a slot. These payments come with an obligation to provide occasional proofs. I saw no reason to treat the period before expiry and the period after expiry different; for a host it means the same effort.

test/Storage.test.js
markspanbroek: I would expect `proofEnd` to change to the current time when the state becomes cancelled or finished. This has the added benefit that `removeEndedContracts()` in `proving.nim` would nicely clean up cancelled or finished contracts.

emizzle:
@markspanbroek what do you think about getting this one merged? Should pave the way for #16 and #19 which are based on top of this one.

contracts/Marketplace.sol
```diff
@@ -178,6 +182,9 @@ contract Marketplace is Collateral, Proofs {
   }
 
   function proofEnd(bytes32 slotId) public view returns (uint256) {
+    if (!_slotAcceptsProofs(slotId)) {
+      return block.timestamp - 1;
+    }
     return _end(slotId);
```
markspanbroek, line 188: If `_end(slotId)` is in the past, I would return that, instead of `block.timestamp - 1`.

markspanbroek (approved):

contracts/Marketplace.sol
emizzle: Updated.

PR #19, opened 2022-09-01
[marketplace] Add "finished" state and extension of proof ending

### LINKED ISSUE #17: Extend proof expectations once all slots are filled
The current logic for expecting proofs (in `expectProofs`) starts when a slot is filled and ends at the `block.timestamp + duration`. This ending will be different for each slot, as each slot will likely be filled at different times, and all slots except the last will not be expected to provide proofs up until the point that the contract is finished. The ending should likely instead be `contractStartTime + duration`. Currently, we don't track `contractStartTime`, so proof expectations will need to be extended (like lock expiry) once all slots are filled.

### LINKED ISSUE #18: Add `isFinished` state calculation

emizzle:
Add function `isFinished` that checks whether the state has been set to `RequestState.Finished` (which is not being set yet) or if the contract was started (`RequestState.Started`) and the contract duration has lapsed.

To facilitate `isFinished`, the contract start time (`startedAt`) was added to calculate the contract end time.

~~Allow proof ending to be extending once a contract is started, so that all filled slots share an ending time that is equal to the contract end time. Added tests.~~

~~Add a mapping for proof id to endId so that proof expectations can be extended for all proofs that share a given endId.~~

The proof end time was modified so that if the request state is not accepting proofs, it takes the min of the slot proof end time, the request end time, or block.timestamp - 1, which ensures that it returns a time in the past. If the slot is accepting proofs, it returns the slot end time.

Add function modifiers that require the request state to allow proofs, with accompanying tests.

General clean up of each function’s request state context, with accompanying tests.

General clean up of all tests, including state change “wait” functions and normalising the time advancement functions.

Closes #17 
Closes #18 

### Notes
1. ~~This branch is based off #16, so probably best to get that merged first, then rebase this on top of main before merging.~~


contracts/Proofs.sol
```diff
@@ -50,17 +64,25 @@ contract Proofs {
...
+    bytes32 id, // typically slot id
+    bytes32 endId, // typically request id, used so that the ending is global for all slots
     uint256 probability,
     uint256 duration
   ) internal {
     require(!ids[id], "Proof id already in use");
     ids[id] = true;
     starts[id] = block.timestamp;
-    ends[id] = block.timestamp + duration;
+    ends[endId] = block.timestamp + duration;
```
markspanbroek, line 82: I wonder whether we shouldn't set the end in a separate function, that we can call from the marketplace once a request for storage is submitted. Then we only have to set it once.

contracts/Proofs.sol
```diff
@@ -161,5 +184,19 @@ contract Proofs {
     missed[id] += 1;
   }
 
+  /// @notice Extends the proof end time
+  /// @dev The id must have a mapping to an end id, the end must exist, and the end must not have elapsed yet
+  /// @param id the id of the proofs to extend. Typically a slot id, the id is mapped to an endId.
+  /// @param ending the new end time (in seconds)
+  function _extendProofEndTo(bytes32 id, uint256 ending) internal {
```
markspanbroek, line 191: Maybe call it `_setProofEnd()`? It could be used to shorten the end time as well, we don't check that the new ending is after the previous ending.

contracts/Storage.sol
```diff
@@ -59,7 +63,12 @@ contract Storage is Collateral, Marketplace {
     return _missed(slotId);
   }
 
-  function isProofRequired(bytes32 slotId) public view returns (bool) {
+  function isProofRequired(bytes32 slotId)
+    public
+    view
+    slotAcceptsProofs(slotId)
```
markspanbroek, line 69: Why would we want to raise an error here when the request is finished?

If we do want an error, then we should probably do the same for `willProofBeRequired()`, `getChallenge()` and `getPointer()`.

markspanbroek (commented):
Thanks @emizzle!

contracts/Marketplace.sol
```diff
@@ -67,12 +68,14 @@ contract Marketplace is Collateral, Proofs {
 
     slot.host = msg.sender;
     slot.requestId = requestId;
-    RequestContext storage context = requestContexts[requestId];
+    RequestContext storage context = _context(requestId);
     context.slotsFilled += 1;
+    context.endsAt = block.timestamp + request.ask.duration;
```
markspanbroek, line 73: A request should end at `block.timestamp + request.ask.duration` in `requestStorage()`, not in`fillSlot()`. The end time of the contract is determined at request submission, it is not dependent on when slots are filled or when the contract is started.

contracts/Marketplace.sol
```diff
@@ -212,11 +232,17 @@ contract Marketplace is Collateral, Proofs {
...
     uint256 end = _end(slotId);
-    if (!_slotAcceptsProofs(slotId)) {
-      return end < block.timestamp ? end : block.timestamp - 1;
+    RequestContext storage context = _context(slot.requestId);
+    if (_slotAcceptsProofs(slotId)) {
+      return end;
+    } else {
+      // Calculate the earliest ending between a slot and a request.
+      // Request endings are set, for eg, when a request fails.
+      uint256 earliestEnd = Math.min(end, context.endsAt);
```
markspanbroek, line 243: There should be no difference between slot end and request end; they should always be the same.

markspanbroek (changes requested):
Nice work @emizzle!
I think it's ready to go once the contract end calculation is fixed.

contracts/Marketplace.sol
emizzle: Each host will need to provide proofs for `request.ask.duration`, correct? This would mean that when the last slot is filled, the request would be active until `block.timestamp + request.ask.duration`.

Otherwise, if the request ended at `time_requested + request.ask.duration`, that means that the host filling the last slot would only need to provide proofs for less than `request.ask.duration`.

markspanbroek: No, hosts only need to provide proofs from the moment they fill a slot until the contract ends. That is indeed shorter for hosts that fill a slot later in time. But they also earn less money because they get paid for the amount of time that they fill a slot.

emizzle: All updated.

emizzle:
@markspanbroek all feedback addressed.

markspanbroek (approved):
Nice! Thanks @emizzle. Let's get this merged :)

PR #22, opened 2022-09-28
List of active requests

markspanbroek:
Maintains a list of active requests per client.
This can be used by the codex node to restore its state when restarted.

Uses an EnumerableSet from OpenZeppelin to ensure that adding and removing from the list is O(1).

Depends on the changes from #20; it needs to be merged first.

### DIFF
--- a/contracts/Marketplace.sol
+++ b/contracts/Marketplace.sol
@@ -267,13 +267,7 @@ contract Marketplace is Collateral, Proofs {
   }
 
   function proofEnd(SlotId slotId) public view returns (uint256) {
-    Slot memory slot = _slot(slotId);
-    uint256 end = _end(_toEndId(slot.requestId));
-    if (_slotAcceptsProofs(slotId)) {
-      return end;
-    } else {
-      return Math.min(end, block.timestamp - 1);
-    }
+    return requestEnd(_slot(slotId).requestId);
   }
 
   function requestEnd(RequestId requestId) public view returns (uint256) {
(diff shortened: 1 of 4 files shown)


contracts/Marketplace.sol
```diff
@@ -30,6 +34,10 @@ contract Marketplace is Collateral, Proofs {
     collateral = _collateral;
   }
 
+  function myRequests() public view returns (RequestId[] memory) {
```
emizzle, line 37: Will this cause an implicit copy from `storage` to `memory`?

markspanbroek: Yes, but that is not so bad; it is a view function that is never called in a transaction.

contracts/Marketplace.sol
```diff
@@ -256,19 +268,23 @@ contract Marketplace is Collateral, Proofs {
 
   function proofEnd(SlotId slotId) public view returns (uint256) {
     Slot memory slot = _slot(slotId);
-    uint256 end = _end(_toEndId(slot.requestId));
+    uint256 end = requestEnd(slot.requestId);
     if (_slotAcceptsProofs(slotId)) {
       return end;
     } else {
       return Math.min(end, block.timestamp - 1);
     }
```
emizzle, line 276: I believe we need to check `_requestAcceptsProofs` as this handles edge cases, eg when a request is cancelled but the state hasn't been updated. The only time proof end is set otherwise is when a request is failed (via `freeSlot`).

markspanbroek: Nice catch!

PR #23, opened 2022-11-04
List active sales

emizzle:
Allows a host to retrieve its active slots. Once a slot is freed or paid out, the active slot is removed. If the request fails, all the slots for the host are removed for that request.

NOTE: While active slots are removed once a slot is freed, if the `maxSlotLoss` is reached and the request is considered failed, the remaining active slots for the host iterated and removed. This requires calling `EnumerableSet.Bytes32.values()` which could potentially run out of gas in the case of an unbounded iteration. However, in this particular case, it should be bounded by the number of slots in the request minus `maxSlotLoss`, which triggers the failure of the request.


markspanbroek:
> While active slots are removed once a slot is freed, if the `maxSlotLoss` is reached and the request is considered failed, the remaining active slots for the host will not be removed.

Would it be correct to phrase the problem as follows? When a request fails, we need to remove all slots from their respective `activeSlots[host]` set. Because we don't know the amount of slots beforehand, this could lead to a loop that has too high a gas cost to be executed.

This seems similar to the problem we had with account locks. There we also need to be able to unlock an unbounded amount of locks when a request finishes. Perhaps we can solve the active slots problem in a similar way that we solved the active locks problem in `AccountLocks`? 

This would entail running a cleanup function (`removeInactiveSlots`) whenever we want to add a new slot to the list. We would also have to ensure that the list never grows to a size whereby this cleanup would take up too much time.
Perhaps we could generalize the code in `AccountLocks` such that we can use it for both account locks and slot lists?

emizzle:
There may not be an issue with an unbounded list as the upper bound will always be the number of slots in the request minus `maxSlotLoss`. The issue is more around the inefficiencies (in gas costs) of listing all active slots (in the `EnumerableSet.Bytes32` data structure) and then removing them. This requires a call to `activeSlots.values` which copies the "entire storage to memory":
>  WARNING: This operation will copy the entire storage to memory, which can be quite expensive. This is designed to mostly be used by view accessors that are queried without any gas fees. Developers should keep in mind that this function has an unbounded cost, and using it as part of a state-changing function may render the function uncallable if the set grows to a point where copying to memory consumes too much gas to fit in a block.

Perhaps the gas consumed may actually be reasonable considering the size of our `EnumerableSet` has a small bound?

Personally, I find that adding an additional layer like a generalised `AccountLocks` may make the code more difficult to reason about. Perhaps there is another solution, such as modifying the `mySlots` function to only return `activeSlots` for a request that isn't failed?

emizzle:
The OpenZeppelin documentation [suggests using an array of `EnumerableSets`](https://docs.openzeppelin.com/contracts/4.x/api/utils#EnumerableSet), and incrementing an index counter as an effective means of clearing it. [This issue in the repo](https://github.com/OpenZeppelin/openzeppelin-contracts/issues/3256#issuecomment-1075670855) points to an example of how to implement it. It turns out that an array panics, so a mapping of `EnumerableSets` was used instead.

emizzle:
I realise now that this removes all slots for all requests, not just the request that failed. Working on a fix...

emizzle:
I reverted the mapping index improvement, and opted to iterate over the values of the `EnumerableSet.Bytes32`. This has the potential to run out of gas in an unbounded iteration, however we are bounded by our `slots - maxSlotLoss`, so it should hopefully not occur. Thoughts?

emizzle:
> If I understand the code correctly, then we're only cleaning up the active slots for one host, not for the other hosts that occupy a slot in the failed request.

Yes, that's correct and it is now changed so that all hosts' slots in that request are removed from `activeSlots`. 

> Users can choose how the value of slots and maxSlotLoss, so they can easily choose values that lead to out of gas problems.

Fair point, I was considering this myself, however my conclusion was that we would be able to limit users from using unreasonable parameters quite easily.

After more thought, and the considerations above, the index mapping for `activeSlots` was reinstated, however the data structure is a bit different. The main difference comes from a change in `mySlots`: it now requires a `RequestId` as a parameter (which can be obtained using `myRequests`. This allowed for the `activeSlots` mapping to be keyed on `RequestId`, which gives a mapping of `ActiveSlotId => EnumerableSet.Bytes32`. `ActiveSlotId` is a keccak hash of `address + activeSlotIdx`, which not only allows us to retrieve all active slot ids per request, but also allows us to clear all active slot ids for a request by incrementing the `activeSlotIdx`.
```
requestId => [ keccak(address + activeSlotIdx) => EnumerableSet.Bytes32 ]
```

contracts/Marketplace.sol
```diff
@@ -7,18 +7,23 @@ import "@openzeppelin/contracts/utils/structs/EnumerableSet.sol";
...
+  type ActiveSlotId is bytes32;
 
   uint256 public immutable collateral;
   MarketplaceFunds private funds;
   mapping(RequestId => Request) private requests;
   mapping(RequestId => RequestContext) private requestContexts;
   mapping(SlotId => Slot) private slots;
   mapping(address => EnumerableSet.Bytes32Set) private activeRequests;
+  mapping(RequestId => mapping(ActiveSlotId => EnumerableSet.Bytes32Set)) private activeSlots;
+  mapping(RequestId => uint8) private activeSlotsIdx;
```
markspanbroek, line 26: Perhaps you could encapsulate `activeSlots` and `activeSlotsId` into a separate contract or struct? Then you can test its logic separately.

markspanbroek (commented):
Nice, I like where you're going with this! Perhaps once this is finished we should consider applying the same approach for `AccountLocks` as well.

> The downside of this is that the caller requires a requestId to call mySlots. This can be obtained using the myRequests() call, but unfortunately adds yet another step to restoring state on the codex side.

`myRequests()` only returns requests that you've submitted yourself. So we'd need to keep track of the requests that host participates in elsewhere. Perhaps we could then also create a getter that read all requests, and for all request their slots. We wouldn't have to worry about gas cost since that would be a view function.

emizzle:
> `myRequests()` only returns requests that you've submitted yourself. So we'd need to keep track of the requests that host participates in elsewhere. Perhaps we could then also create a getter that read all requests, and for all request their slots. We wouldn't have to worry about gas cost since that would be a view function.

When restoring the state in a restarted node, we'd only need to get requests that we've submitted ourselves, no?

emizzle:
The large mapping piece was moved to a separate library (without tests yet) with a generalised API. I'm not entirely sure how we can use this library to replace `AccountLocks`, as there is an expiry piece that won't be handled.

I could be mistaken, but I feel that `AccountLocks` could be replaced by `activeRequests` which is a mapping of `address => EnumerableSet.Bytes32 (RequestId)`. Collateral could simply require there are no requests for an address, or no requests that aren't expired, before allowing withdrawal. Maybe this is what you meant by "perhaps we could then also create a getter that read all requests, and for all request their slots". This would be the new `activeRequests`, replacing `AccountLocks`. `Collateral` doesn't have access to `activeRequests` however, because it's a property of `Marketplace` and `Marketplace` inherits from `Collateral`.

emizzle:
Latest commit adds ability to get all requests, which will be nice to list all requests > addresses > slots. However, this still doesn't work for AccountLocks because of inheritance. Ideas?

markspanbroek:
> When restoring the state in a restarted node, we'd only need to get requests that we've submitted ourselves, no?

The Purchasing module would need to get all requests that it submitted itself. The Sales module would need to get all requests where it filled a slot.

> The large mapping piece was moved to a separate library

Very nice!

> I could be mistaken, but I feel that `AccountLocks` could be replaced by `activeRequests`

Good idea, that could simplify things a lot.

> `Collateral` doesn't have access to `activeRequests`

Collateral only needs to check whether a withdrawal is allowed.  We could make Collateral abstract, and define a virtual function in Collateral that Marketplace could then implement:

```solidity
function _isWithdrawAllowed(address account) internal virtual view returns (bool);
```

But it would be perfectly ok to leave the rework of AccountLocks for a later PR.

contracts/Marketplace.sol
```diff
@@ -41,19 +43,50 @@ contract Marketplace is Collateral, Proofs {
...
+    return _toRequestIds(Utils._resize(_toBytes32s(result), counter));
+  }
+
+  function requestsForHost(address host) public view returns(RequestId[] memory) {
+    EnumerableSet.Bytes32Set storage keys = activeRequestsForHosts.keys();
+    uint256 keyLength = keys.length();
+    RequestId[] memory result = new RequestId[](keyLength);
+
+    uint8 counter = 0; // should be big enough
+    for (uint8 i = 0; i < keyLength; i++) {
```
markspanbroek, line 65: Doesn't this iterate through all active requests in the entire network? Even for a view function that seems excessive.
The index `i` is not going to fit in a `uint8` :smile:

contracts/Marketplace.sol
```diff
@@ -41,19 +43,50 @@ contract Marketplace is Collateral, Proofs {
...
+    }
+    return _toRequestIds(Utils._resize(_toBytes32s(result), counter));
+  }
+
+  function requestsForHost(address host) public view returns(RequestId[] memory) {
+    EnumerableSet.Bytes32Set storage keys = activeRequestsForHosts.keys();
+    uint256 keyLength = keys.length();
+    RequestId[] memory result = new RequestId[](keyLength);
+
+    uint8 counter = 0; // should be big enough
```
markspanbroek, line 64: 256 hosts per contract seems conservative, why not use a bigger int?

contracts/Marketplace.sol
emizzle: There is a large trade off between being able to clear all addresses for a request (once a request is cancelled/failed) and being able to view all requests for an address. 

I previously tried using an `AddressBytes32SetMap` which maps `address => RequestIds`. This allows for an efficient view of requests for an address (`requestsForHosts`), however, as mentioned, removing a request once cancelled/failed, requires iteration in a transaction. We could take this a step further by implementing filtering of `RequestIds` based on state inside `requestsForHosts`, which should accurately show requests for a host. However cancelled/failed `RequestIds` would never be removed, thus making the iteration longer and longer as time passes. I was thinking perhaps that we could go even further, doing something similar to `AccountLocks` by removing cancelled/failed requests on each `add`, however, again this requires iteration in a transaction.

Looking at other options...

contracts/Marketplace.sol
```diff
@@ -143,12 +210,20 @@ contract Marketplace is Collateral, Proofs {
   {
     require(_isFinished(requestId), "Contract not ended");
     RequestContext storage context = _context(requestId);
-    Request storage request = _request(requestId);
+    // Request storage request = _request(requestId);
```
markspanbroek, line 213:
```suggestion
```

contracts/libs/Mappings.sol
```diff
@@ -0,0 +1,232 @@
+// SPDX-License-Identifier: MIT
+// heavily inspired by: https://bitbucket.org/rhitchens2/soliditystoragepatterns/src/master/GeneralizedCollection.sol
+pragma solidity ^0.8.8;
+
+import "./Debug.sol"; // DELETE ME
```
markspanbroek, line 5: Remove this?

test/SetMap.test.js
```diff
@@ -0,0 +1,207 @@
+const { ethers } = require("hardhat")
```
markspanbroek, line 1: This file can be deleted as well

contracts/libs/Mappings.sol
```diff
@@ -0,0 +1,232 @@
...
+
+library Mappings {
+  type KeyId is bytes32;
+  type ValueId is bytes32;
+
+  // first entity is called a "One"
```
markspanbroek, line 11: Might as well remove these comments; they are no longer up to date, as these type of comments tend to do ;)

contracts/libs/Mappings.sol
```diff
@@ -0,0 +1,232 @@
+// SPDX-License-Identifier: MIT
+// heavily inspired by: https://bitbucket.org/rhitchens2/soliditystoragepatterns/src/master/GeneralizedCollection.sol
```
markspanbroek, line 2: I think you mean: https://bitbucket.org/rhitchens2/soliditystoragepatterns/src/master/OneToMany.sol
It does not have a license, so we cannot just use it.

contracts/libs/Mappings.sol
markspanbroek, line 5: Remove?

contracts/Marketplace.sol
```diff
@@ -35,7 +46,43 @@ contract Marketplace is Collateral, Proofs {
...
+    uint256 counter = 0;
+    uint256 totalSlots = activeRequestSlots.getManyCount(); // set this bigger than our possible filtered list size
+    if (totalSlots == 0) {
+      return new SlotId[](0);
+    }
+    bytes32[] memory result = new bytes32[](totalSlots);
+    Mappings.ValueId[] storage valueIds =
+      activeHostRequests.getValueIds(Mappings.toKeyId(msg.sender));
+    for (uint256 i = 0; i < valueIds.length; i++) {
+      // There may exist slots that are still "active", but are part of a request
```
markspanbroek, line 68: This comment is no longer up to date?

contracts/Marketplace.sol
```diff
@@ -35,7 +46,43 @@ contract Marketplace is Collateral, Proofs {
...
+    public
+    view
+    returns (SlotId[] memory)
+  {
+    uint256 counter = 0;
+    uint256 totalSlots = activeRequestSlots.getManyCount(); // set this bigger than our possible filtered list size
+    if (totalSlots == 0) {
+      return new SlotId[](0);
+    }
+    bytes32[] memory result = new bytes32[](totalSlots);
```
markspanbroek, line 64: Memory usage of this function will grow with the total amount of slots in the storage network. That could lead to problems down the road.

contracts/Marketplace.sol
```diff
@@ -35,7 +46,43 @@ contract Marketplace is Collateral, Proofs {
...
+  }
+
+  function mySlots()
+    public
+    view
+    returns (SlotId[] memory)
+  {
+    uint256 counter = 0;
+    uint256 totalSlots = activeRequestSlots.getManyCount(); // set this bigger than our possible filtered list size
+    if (totalSlots == 0) {
```
markspanbroek, line 61: I would avoid this optimization. It will only be used when there has been no activity in the network yet.

markspanbroek (commented):
I'll have to study the one-to-many solution a bit more, but here's my initial comments.

markspanbroek:
Closing because alternative #24 was merged.

PR #28, opened 2023-01-10
Simplify proof start and end

markspanbroek:
The timestamps when requests started and ended were stored in both the `Marketplace` and `Proofs`. This PR simplifies this by storing the request start and end only in the Marketplace. `Proofs` access them through virtual functions that are implemented by the `Marketplace`.


contracts/Marketplace.sol
```diff
@@ -284,28 +282,18 @@ contract Marketplace is Collateral, Proofs {
...
   function getHost(SlotId slotId) public view returns (address) {
-    return _host(slotId);
+    return slots[slotId].host;
   }
 
-  function _request(
+  function getRequest(
     RequestId requestId
-  ) internal view returns (Request storage) {
+  ) public view returns (Request memory) {
```
emizzle, line 291: This creates a storage > memory copy, right? Any particular reason this is needed?

emizzle (approved):
Great job mark, was going to suggest similar improvements.

One question regarding the `getRequest` change which I believe introduces a storage to memory. What's the benefit here?

contracts/Marketplace.sol
markspanbroek: I wanted to combine `_request()` and `getRequest()`, but since `getRequest` is public, it is not allowed to return a storage location. Now that you mention it, we should probably keep the two separate. I'll revert this change.

PR #29, opened 2023-01-10
Introduce abstraction for dealing with time periods

markspanbroek:
Extracts the logic from `Proofs` that deals with time periods into `Periods.sol`.

### DIFF
--- a/contracts/Marketplace.sol
+++ b/contracts/Marketplace.sol
@@ -129,7 +129,7 @@ contract Marketplace is Collateral, Proofs {
 
   function markProofAsMissing(
     SlotId slotId,
-    uint256 period
+    Period period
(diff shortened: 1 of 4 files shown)


emizzle (approved):
LGTM. `Periods` could also be a `library` instead of a `contract`, however `secondPerPeriod` would not be storable in the lib, instead it would need to be passed to `periodOf`.

markspanbroek:
> `Periods` could also be a `library` instead of a `contract`, however `secondPerPeriod` would not be storable in the lib, instead it would need to be passed to `periodOf`

Yes, and also `periodStart()` and `periodEnd()` would need to have it passed to them. That's why I choose to make it a `contract` for now.

PR #31, opened 2023-01-11
Fix: use slot start instead of request start

markspanbroek:
Proofs should be required from the moment a slot is filled, not from the moment a request is submitted.

Reverts part of #28.


emizzle (approved):
Great catch! This is much cleaner than using the virtual function, and it makes sense to keep track of proof start times inside of `Proofs`. The other option is to keep track of slot starts inside of `Marketplace`, but it's not needed for anything inside of `Marketplace`, so this approach is cleaner.

PR #33, opened 2023-01-18
Fix proof probability

markspanbroek:
* fixes the case where the probability is 1 (proofs always required)
* uses a binomial test to check probabilities (more reliable unit test)

### DIFF
--- a/test/Proofs.test.js
+++ b/test/Proofs.test.js
@@ -12,6 +12,7 @@ const {
 } = require("./evm")
 const { periodic } = require("./time")
 const { SlotState } = require("./requests")
+const binomialTest = require("@stdlib/stats-binomial-test")
 
 describe("Proofs", function () {
   const slotId = hexlify(randomBytes(32))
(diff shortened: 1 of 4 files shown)


test/Proofs.test.js
```diff
@@ -40,16 +41,32 @@ describe("Proofs", function () {
...
         if (await proofs.isProofRequired(slotId)) {
           amount += 1
         }
         await advanceTime(period)
       }
-      let expected = 100 / probability
-      expect(amount).to.be.closeTo(expected, expected / 2)
+
+      const p = 1 / probability // expected probability
+      const alpha = 1 / 1000 // unit test can fail once every 1000 runs
```
emizzle, line 57: Please excuse my lack of understanding of proof probabilities. Why are we checking a probability? Why are there allowed failures? In my mind, it seems to make sense that that a node will need to provide a proof every period, that's it. Adding in a probability seems to make it such that there is a chance a node won't need to provide a proof during the period.

markspanbroek: The problem with testing probabilities involving random sampling is that it is non-deterministic. In this unit test I try to strike a balance between finding errors in the implementation and not having the test fail because of a couple of unlucky throws of the dice.

emizzle: Makes sense, thanks Mark.

emizzle (approved):

Issue #43, opened 2023-02-21
Canceled request's slots payouts money to Hosts


AuHau:
When a Request is canceled because of time out (after the Client calls the `withdrawFunction`, which returns him all the money for the original Request), then hosts that have already filled some slots can call `freeSlot` which pays them out the money for the slot. This makes the contract's money pool out of balance as the money is already returned to the Client in the `withdrawFunction`.

There is a plan in the future to take into consideration the time for how long was the content hosted and return only a partial reward to the client and a partial reward for the hosts, but until then this needs to be fixed.

The probably easiest solution is to payout filled slots to host for now.

AuHau:
Fixed with https://github.com/codex-storage/codex-contracts-eth/pull/69

PR #37, opened 2023-01-30
Use fuzzing to check invariants

markspanbroek:
Simplifies the contract code by moving the invariant checks to separate files.
Uses [echidna](https://github.com/crytic/echidna) to perform the fuzzing.

PR #44, opened 2023-02-21
feat: collateral per slot

### LINKED ISSUE #38: Collateral per slot
Currently, a host puts down collateral once, which allows it to fill slots in many storage contracts. We'd like to change that to a (smaller) collateral per slot that the host fills. That collateral is then specifically meant to cover risks associated with that slot.

Rationale and design:
https://github.com/status-im/codex-research/blob/main/design/marketplace.md

AuHau:
The code is ready for review, but I have not yet touched the tests. Some remarks below.

I have removed the `Collateral` abstract contract. The collateral is now pretty much tied to the `Slot` structure, and I don't see clearly how it could be split into a separate contract like it was till now. But if you think there is a way, I am happy to hear it! ;-) This means, though, that the collateral funds tracking and `collateralInvariant` is removed, but as we want to move toward fuzzing anyhow, I don't think it is necessary

Regarding timed-out Requests. I have considered for the Client to return the collaterals for already filled slots (in the `withdrawFunds()` method), but since `uint64` can be quite a big number of slots to cover, I have decided to leave it up to the Host to call the `freeSlot()` at his discretion when he detects that the Request of his Slot was canceled and wants his collateral back. This could be improved by allowing calling this in bulk. Currently, as discussed in #43, the hosts also get the slot's reward, but in the follow-up PR, I will change that for them to get only the collateral back.

Closes #38


contracts/Marketplace.sol
```diff
@@ -120,11 +130,13 @@ contract Marketplace is Collateral, Proofs, StateRetrieval {
   function markProofAsMissing(SlotId slotId, Period period) public {
     require(slotState(slotId) == SlotState.Filled, "Slot not accepting proofs");
     _markProofAsMissing(slotId, period);
-    address host = getHost(slotId);
+    Slot storage slot = _slots[slotId];
+
     if (missingProofs(slotId) % config.collateral.slashCriterion == 0) {
-      _slash(host, config.collateral.slashPercentage);
+      uint256 slashedAmount = (slot.currentCollateral * config.collateral.slashPercentage) / 100;
+      slot.currentCollateral -= slashedAmount;
```
markspanbroek, line 137: You'd probably want to update `_funds.slashed` as well here

contracts/Marketplace.sol
```diff
@@ -212,10 +215,6 @@ contract Marketplace is Collateral, Proofs, StateRetrieval {
     require(token.transfer(msg.sender, amount), "Withdraw failed");
   }
 
-  function getHost(SlotId slotId) public view returns (address) {
```
markspanbroek, line 215: We do need this for other purposes (checking whether a host filled a slot or another host got there first)

contracts/Configuration.sol
```diff
@@ -9,7 +9,7 @@ struct MarketplaceConfig {
 }
 
 struct CollateralConfig {
-  uint256 initialAmount; // amount of collateral necessary to fill a slot
+  uint256 minimalInitialAmount; // amount of collateral necessary to fill a slot
```
markspanbroek, line 12: I don't think that we need this anymore; a client is free to choose whatever amount of collateral they wish (even zero)

markspanbroek (commented):
Nicely done, thanks @AuHau! I left a few small comments.

contracts/Marketplace.sol
AuHau: As I mentioned in the PR's description, I have removed the collateral funds tracking as part of the Collateral contract removal. The `_funds` in play here are the `MarketplaceFunds` that do not have `slashed` property. Do you want me to incorporate that? Or will this be removed with the fuzzing anyhow?

contracts/Configuration.sol
AuHau: Yeah, I was considerating removing it, but then realized that when the repair comes into game, then there should be always some collateral requested so the funds could be used for funding the repair?

contracts/Marketplace.sol
markspanbroek: The implementation will be different when #37 (fuzzing) is merged, but I'd still like to check that slashed funds can never be extracted from the contract. Probably the easiest way of doing that now is by keeping track of slashed funds in the Marketplace, since we no longer need the Collateral contract.

contracts/Marketplace.sol
```diff
@@ -6,20 +6,20 @@ import "@openzeppelin/contracts/utils/math/Math.sol";
...
   using Requests for Request;
 
+  IERC20 public immutable token;
   MarketplaceConfig public config;
 
   MarketplaceFunds private _funds;
   mapping(RequestId => Request) private _requests;
   mapping(RequestId => RequestContext) private _requestContexts;
-  mapping(SlotId => Slot) private _slots;
+  mapping(SlotId => Slot) internal _slots;
```
AuHau, line 22: I have changed this in order to have testing functions related to this in the `TestMarketplace`

contracts/Marketplace.sol
```diff
@@ -212,10 +211,6 @@ contract Marketplace is Collateral, Proofs, StateRetrieval {
     require(token.transfer(msg.sender, amount), "Withdraw failed");
   }
 
-  function getHost(SlotId slotId) public view returns (address) {
```
AuHau, line 215: I have moved this function into the `TestCollateral`

markspanbroek: Well, we actually need this in the Codex client, so moving it to TestCollateral won't work :smile: The sales module in nim-codex relies on being able to see which host filled a slot.

contracts/Configuration.sol
```diff
@@ -9,7 +9,6 @@ struct MarketplaceConfig {
 }
 
 struct CollateralConfig {
-  uint256 initialAmount; // amount of collateral necessary to fill a slot
   uint256 minimumAmount; // frees slot when collateral drops below this minimum
   uint256 slashCriterion; // amount of proofs missed that lead to slashing
   uint256 slashPercentage; // percentage of the collateral that is slashed
```
markspanbroek, line 14: Thinking about this a bit more, we probably want to move these other collateral related constants into the `Request` struct as well.

We don't want to enforce a network-wide minimum amount of collateral, so moving `minimumAmount` to `Request` makes sense. As for `slashCriterion` and `slashPercentage`, we could go either way with these. They could either be network-wide constants, or set per request.

markspanbroek (commented):
Thanks @AuHau, I like the tests! And good that you caught the missing editorconfig.

contracts/Marketplace.sol
AuHau: Oh, damn 🙈 I will revert it then.

markspanbroek (approved):
Awesome, looks good!

PR #47, opened 2023-03-06
feat: collateral fractions

### LINKED ISSUE #46: Repair reward percentage, number of slashing and linearity
Currently, the minimum required collateral is defined on the smart-contract level as an absolute value. Repair reward is currently not defined. And the slashing is logarithmic.

We have agreed to replace the `minimumAmount` with number of slashing. There needs to be a check for `numberOfSlashing*slashingPercentage<=100`, though. Also we have decided to move from logarithmic to linear slashing, which goes in hand with the previous change.

AuHau:
Closes #46 

Also does a bit of space optimization on the used numeric types for the config values.

### DIFF
--- a/contracts/Marketplace.sol
+++ b/contracts/Marketplace.sol
@@ -139,9 +139,8 @@ contract Marketplace is Proofs, StateRetrieval {
       _funds.balance -= slashedAmount;
 
       if (missingProofs(slotId) / config.collateral.slashCriterion >= config.collateral.maxNumberOfSlashes) {
-        // When the collateral drops below the minimum threshold, the slot
-        // needs to be freed so that there is enough remaining collateral to be
-        // distributed for repairs and rewards (with any leftover to be burnt).
+        // When the number of slashings is at or above the allowed amount,
+        // free the slot.
         _forciblyFreeSlot(slotId);
       }
     }
--- a/contracts/Marketplace.sol
+++ b/contracts/Marketplace.sol
@@ -45,9 +45,9 @@ contract Marketplace is Proofs, StateRetrieval {
   ) Proofs(configuration.proofs) marketplaceInvariant {
     token = token_;
 
-    require(configuration.collateral.minimumAmountPercentage <= 100, "Too big percentage");
-    require(configuration.collateral.repairRewardPercentage <= 100, "Too big percentage");
-    require(configuration.collateral.slashPercentage <= 100, "Too big percentage");
+    require(configuration.collateral.repairRewardPercentage <= 100, "Must be less than 100");
+    require(configuration.collateral.slashPercentage <= 100, "Must be less than 100");
+    require(configuration.collateral.maxNumberOfSlashes * configuration.collateral.slashPercentage <= 100, "Total slash percentage must be less then 100");
     config = configuration;
   }
 
@@ -133,12 +133,12 @@ contract Marketplace is Proofs, StateRetrieval {
     Request storage request = _requests[slot.requestId];
 
     if (missingProofs(slotId) % config.collateral.slashCriterion == 0) {
-      uint256 slashedAmount = (slot.currentCollateral * config.collateral.slashPercentage) / 100;
+      uint256 slashedAmount = (request.ask.collateral * config.collateral.slashPercentage) / 100;
       slot.currentCollateral -= slashedAmount;
       _funds.slashed += slashedAmount;
       _funds.balance -= slashedAmount;
 
-      if (slot.currentCollateral < minimumAmount(request)) {
+      if (missingProofs(slotId) / config.collateral.slashCriterion >= config.collateral.maxNumberOfSlashes) {
         // When the collateral drops below the minimum threshold, the slot
         // needs to be freed so that there is enough remaining collateral to be
         // distributed for repairs and rewards (with any leftover to be burnt).
@@ -295,10 +295,6 @@ contract Marketplace is Proofs, StateRetrieval {
       return SlotState.Failed;
     }
     return slot.state;
-  }
-
-  function minimumAmount(Request memory request) private view returns(uint256) {
-    return (request.ask.collateral * config.collateral.minimumAmountPercentage ) / 100;
   }
 
   function _transferFrom(address sender, uint256 amount) internal {
--- a/contracts/Marketplace.sol
+++ b/contracts/Marketplace.sol
@@ -44,6 +44,10 @@ contract Marketplace is Proofs, StateRetrieval {
     MarketplaceConfig memory configuration
   ) Proofs(configuration.proofs) marketplaceInvariant {
     token = token_;
+
+    require(configuration.collateral.minimumAmountPercentage <= 100, "Too big percentage");
+    require(configuration.collateral.repairRewardPercentage <= 100, "Too big percentage");
+    require(configuration.collateral.slashPercentage <= 100, "Too big percentage");
     config = configuration;
   }
 
@@ -126,6 +130,7 @@ contract Marketplace is Proofs, StateRetrieval {
     require(slotState(slotId) == SlotState.Filled, "Slot not accepting proofs");
     _markProofAsMissing(slotId, period);
     Slot storage slot = _slots[slotId];
+    Request storage request = _requests[slot.requestId];
 
     if (missingProofs(slotId) % config.collateral.slashCriterion == 0) {
       uint256 slashedAmount = (slot.currentCollateral * config.collateral.slashPercentage) / 100;
@@ -133,7 +138,7 @@ contract Marketplace is Proofs, StateRetrieval {
       _funds.slashed += slashedAmount;
       _funds.balance -= slashedAmount;
 
-      if (slot.currentCollateral < config.collateral.minimumAmount) {
+      if (slot.currentCollateral < minimumAmount(request)) {
         // When the collateral drops below the minimum threshold, the slot
         // needs to be freed so that there is enough remaining collateral to be
         // distributed for repairs and rewards (with any leftover to be burnt).
@@ -290,6 +295,10 @@ contract Marketplace is Proofs, StateRetrieval {
       return SlotState.Failed;
     }
     return slot.state;
+  }
+
+  function minimumAmount(Request memory request) private view returns(uint256) {
+    return (request.ask.collateral / 100) * config.collateral.minimumAmountPercentage;
   }
 
   function _transferFrom(address sender, uint256 amount) internal {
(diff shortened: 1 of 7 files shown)


AuHau:
I have not tested underflow as ABI typings checks handle that. Should I add tests for that as well?

contracts/Marketplace.sol
```diff
@@ -44,6 +44,10 @@ contract Marketplace is Proofs, StateRetrieval {
     MarketplaceConfig memory configuration
   ) Proofs(configuration.proofs) marketplaceInvariant {
     token = token_;
+
+    require(configuration.collateral.minimumAmountPercentage <= 100, "Too big percentage");
```
emizzle, line 48: Might be more informative:
```suggestion
    require(configuration.collateral.minimumAmountPercentage <= 100, "Must be less than 100");
```

emizzle (commented):
After listening to last week's call and seeing the implementation, it has got me thinking that perhaps we could calculate the reward instead of specifying it as a network-wide parameter.

The maximum repair reward could be:
1. Total slashings from slot +
2. Total proof payouts for slot up to the point of being freed

This is the maximum that could given out, so there's enough remaining to payout future slot fills. If we want to decrease this amount (to disincentivise too much reward for repairing a slot), perhaps we could introduce a parameter to decrease the repair reward by a percentage, and burn it.

To facilitate this, we could also make each slashing amount be equal to the `request.ask.collateral * config.collateral.slashPercentage`, such that `currentCollateral` is linearly decreased, instead of logarithmically. This would allow us to specify a whole number `slashingsAllowed` parameter, instead of a collateral amount or percentage, and make it easier to reason about required parameters and reason in the code.

In summary, it is being suggested to:
1. replace `repairRewardPercentage` with `slashingsAllowed`
3. in the constructor, validate that `slashingsAllowed > 100 / slashPercentage` (to constrain the related parameters)
3. Change `slashedAmount` to be `request.ask.collateral * config.collateral.slashPercentage`
4. track slot slashings and change the logic for freeing a slot to be: `if (slot.slashings >= slashingsAllowed) { _forciblyFreeSlot(slotId) }`

Thoughts?

markspanbroek:
@AuHau:
> I have not tested underflow as ABI typings checks handle that. Should I add tests for that as well?

No, I don't think we don't need tests for that.

@emizzle:
> The maximum repair reward could be:
> 
>     1. Total slashings from slot +
>     2. Total proof payouts for slot up to the point of being freed

This would mean that the repair reward fluctuates over time, which I would not be in favor of because it makes it more difficult to reason about. Also, as a design guideline we've been trying to keep rewards as close as possible to costs, to make sure that we don't introduce unwanted incentives.

>  we could also make each slashing amount be equal to the `request.ask.collateral * config.collateral.slashPercentage`, such that `currentCollateral` is linearly decreased, instead of logarithmically

That is a good idea; would make slashing more simple to reason about.

markspanbroek (approved):

emizzle:
> This would mean that the repair reward fluctuates over time, which I would not be in favor of because it makes it more difficult to reason about

That means then we have to consider using the maximum repair reward + payouts for constraints / conditions.

The amount of collateral we have at any given time must be greater than `rewardRepairPercentage * request.ask.collateral` plus all potential payouts owing `pricePerSlot(request)`. 
 Therefore, we don't need `minimumAmountPercentage` at all, because that *value* can be calculated: `rewardRepairPercentage * request.ask.collateral + pricePerSlot(request)`.

Meaning, we could slash first, then check our conditions for freeing a slot:
```solidity
uint256 rewardsMax = repairRewardPercentage * request.ask.collateral + pricePerSlot(request);
if (slot.currentCollateral <= rewardsMax) {
          _forciblyFreeSlot(slotId)
}
```
(I'll retract my previous suggestion about using `slashingsAllowed` instead of `slashPercentage`, as this would create another condition in the `if` statement above to check for, because specifying a number of slashings may not coincide with the collateral amount.)

Personally, I'm trying to limit the number of parameters and amount of confusion a user would encounter when specifying parameters while still giving full flexibility.  As above, a user would *only* need to specify `repairRewardPercentage` and `slashPercentage`, and not `minimumAmountPercentage`. If `minimumAmountPercentage` were kept, then a constraint would need to be enforced such that:
```solidity
uint256 minCollateral = minimumAmountPercentage * request.ask.collateral;
uint256 rewardsMax = repairRewardPercentage * request.ask.collateral + pricePerSlot(request);
require(minCollateral > rewardsMax, "minimumAmountPercentage too high and/or rewardRepairPercentage too low"); // I know this msg is too long
```
But in my opinion, as a user it may not be clear what `minimumAmountPercentage` is and how it relates to `repairRewardPercentage` and `slashPercentage`.

Sorry to drag this out further, but another thing we may want to consider is *not* slashing first, but checking that if we were to slash, our slot collateral would drop below the `rewardsMax`, and therefore free the slot. I had this implemented at one point, but we reverted it in favour of slashing first, and I can't remember why. But doing the check first, would ensure that we always have enough funds (unless maybe if we decide to take funds from the slashed funds 🤔).

AuHau:
@emizzle I think there might be some confusion here ;-) 

> Personally, I'm trying to limit the number of parameters and amount of confusion a user would encounter when specifying parameters while still giving full flexibility

I completely agree with your opinion that we should keep things simple for the user and not expose him to too many parameters that he has to set and decide. Actually, that was the reason for this PR, as Mark originally suggested in the [per-slot collateral PR](https://github.com/status-im/codex-contracts-eth/pull/44#discussion_r1121991096) to have these parameters as part of the `Request.ask` structure and I argued against this for this exact reason. And this PR does exactly that by taking those percentages to the contract's constructor level (or to the "network level"). Therefore the user does not need to specify those, and so I am a bit confused why you say that he has to. Or are you arguing against having those on the constructor level?

> The amount of collateral we have at any given time must be greater than rewardRepairPercentage * request.ask.collateral plus all potential payouts owing pricePerSlot(request).
Therefore, we don't need minimumAmountPercentage at all, because that value can be calculated: rewardRepairPercentage * request.ask.collateral + pricePerSlot(request).

With this, you are saying to split the collateral into two parts. One gets slashed, and one is used for the repair reward. Right? To my understanding, there is a feature in the current design that splits the collateral into three parts, because we want to have control over the repair reward that gets paid out in order not to be too big to create some potential attacks.

That said, I am not sure if using fractions of the collateral will achieve it? As in the scope of this PR, the reward can be pretty big if the collateral ask was also big. But then, I guess that is alright because big collateral should signal the importance of the data and hence bigger collateral ask will most likely result in a premium price for the Request and hence bigger repair reward should lead to a better incentive for somebody to come in to repair the important data.

> 2. Total proof payouts for slot up to the point of being freed

I was originally under the assumption that this is the case and it should be part of the repair reward, but Marked explained to me that it is not the case [here](https://github.com/status-im/codex-research/pull/160#discussion_r1129273881). We also discussed it later on a call where he correctly pointed out that this could create an incentive for bad actors to DDOS nodes that are about to finish a Request in order to get their money, which I agree is indeed potentially a big problem.

Sooo, maybe we should jump on a call on Monday and discuss it further if you see the need?

emizzle:
>  Therefore the user does not need to specify those, and so I am a bit confused why you say that he has to. Or are you arguing against having those on the constructor level?

My apologies for the confusion Adam! 😅 You are right, this is already at the network level, my mistake.

I think the overall goal I was hoping to achieve with my suggestions was to keep the parameters at a minimum *if we can*. One reason is to keep user-facing complexity at a minimum, and the other is to keep the code easy to reason about. Therefore, I thought maybe we could attempt to reduce the parameter count if we didn't *need* all of them (and if we do need all of them, then that's completely ok too).

[When I first introduced the concept of the minimum collateral needed](https://github.com/status-im/codex-contracts-eth/pull/16/files#diff-e4c1cadd00d0408c993a439f87779b0eb6f035eef31fe4c1c47a3d7c7b1e0a39R89-R97), personally I was kind of intending it to be a stop gap for not having the repair functionality / costing concretely defined yet. In other words, I assumed at some point we'd change the parameter to align with repair. And it does seem that we have done that with `rewardRepairPercentage` to a degree (more on this below), so it has made me then wonder if we need BOTH `minimumAmountPercentage` AND `rewardRepairPercentage`, or would it be possible to go with one or the other. In [my second comment above](https://github.com/status-im/codex-contracts-eth/pull/47#issuecomment-1463271283), I think I started to realise that if we have decided that `rewardRepairPercentage` is good enough, then we don't really need `minimumAmountPercentage` as it could be calculated.

> That said, I am not sure if using fractions of the collateral will achieve it? As in the scope of this PR, the reward can be pretty big if the collateral ask was also big. But then, I guess that is alright because big collateral should signal the importance of the data and hence bigger collateral ask will most likely result in a premium price for the Request and hence bigger repair reward should lead to a better incentive for somebody to come in to repair the important data.

I completely agree that it makes sense for the repair reward to grow linearly with the size of the collateral, however, as you alluded to, this would likely deviate from the actual repair costings if the collateral is sufficiently large, which could incentivise an attack vector for large collateral contracts. @markspanbroek please correct me if I'm wrong, but we won't have an accurate idea of how to estimate repair costs until we have bandwidth incentives worked out, correct? If that's the case, perhaps we should leave the repair reward as a percentage of collateral for now until we work out bandwidth incentives and a more accurate way to represent repair rewards as a function of costings?

emizzle:
I'd also love to jump on a call today to discuss ☺️

AuHau:
As discussed, I have implemented the `maxNumberOfSlashes` and linear decrease of the collateral.

markspanbroek (approved):

contracts/Marketplace.sol
```diff
@@ -133,12 +133,12 @@ contract Marketplace is Proofs, StateRetrieval {
...
+      uint256 slashedAmount = (request.ask.collateral * config.collateral.slashPercentage) / 100;
       slot.currentCollateral -= slashedAmount;
       _funds.slashed += slashedAmount;
       _funds.balance -= slashedAmount;
 
-      if (slot.currentCollateral < minimumAmount(request)) {
+      if (missingProofs(slotId) / config.collateral.slashCriterion >= config.collateral.maxNumberOfSlashes) {
         // When the collateral drops below the minimum threshold, the slot
         // needs to be freed so that there is enough remaining collateral to be
         // distributed for repairs and rewards (with any leftover to be burnt).
```
emizzle, line 144:
```suggestion
        // When the number of slashings is at or above the allowed amount,
        // free the slot.
```

emizzle (approved):
Thanks Adam, looks great! Just a couple of minor comments updates if you don't mind. Then looks good to merge 😊

PR #48, opened 2023-03-06
fix: reset missed counter when slot is freed

AuHau:
When a slot is freed after the host misses too many proofs, the "missed proofs counter" is not reset, and hence the new host that fills the freed slot will have a non-zero "missed proofs counter". While this does not impact functionality as the logic uses modulo in the implementation, if somebody queries the `missingProofs()` method directly, he would be surprised with the number he would potentially get.


markspanbroek (approved):
Good find!

PR #51, opened 2023-03-29
retrieve active slot for sales state restoration

emizzle:
Store slotIndex in slot struct and expose the slot via getActiveSlot. This is to be used when restoring state in the sales module after a node restart.


AuHau (approved):
LGTM

PR #52, opened 2023-03-30
build: funds localhost env. accounts

AuHau:
This PR adds funding of the TestToken to development accounts

### DIFF
--- a/deploy/token.js
+++ b/deploy/token.js
@@ -5,7 +5,7 @@ module.exports = async ({ deployments, getNamedAccounts, getUnnamedAccounts }) =
   const tokenDeployment = await deployments.deploy("TestToken", { from: deployer })
   const token = await hre.ethers.getContractAt("TestToken", tokenDeployment.address)
 
-  const accounts = await getUnnamedAccounts()
+  const accounts = [...Object.values(await getNamedAccounts()), ...(await getUnnamedAccounts())]
   for (const account of accounts) {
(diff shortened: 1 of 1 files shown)


deploy/token.js
```diff
@@ -1,6 +1,17 @@
-module.exports = async ({ deployments, getNamedAccounts }) => {
+const MINTED_TOKENS = 1_000_000_000
+
+module.exports = async ({ deployments, getNamedAccounts, getUnnamedAccounts }) => {
   const { deployer } = await getNamedAccounts()
-  await deployments.deploy("TestToken", { from: deployer })
+  const tokenDeployment = await deployments.deploy("TestToken", { from: deployer })
+  const token = await hre.ethers.getContractAt("TestToken", tokenDeployment.address)
+
+  const accounts = await getUnnamedAccounts()
```
markspanbroek, line 8: Looks like it doesn't mint test tokens for named accounts (account #0 in this case). Is that on purpose? I was a bit surprised that account #0 didn't get any coins.

markspanbroek (approved):
Looks good!

deploy/token.js
AuHau: Yeah it surprised me as well when I was testing it yesterday 😅 We can also mint for it, I don't really mind. Maybe though as it is deployer we should not really use it for testing?

AuHau: I have added also the named accouns in order to prevent possible confusions ;-)

PR #54, opened 2023-04-21
build: make sure marketplace is on predefined address

AuHau:
Needed for https://github.com/status-im/nim-codex/issues/372

It is not needed per-se as it seems that the address should be deterministic, but if there would be a case when we add a transaction into the deployment script, then the address would change, so this is just a safe guard to make it future-proof.

### DIFF
--- a/deploy/marketplace.js
+++ b/deploy/marketplace.js
@@ -31,7 +31,6 @@ async function aliasContract({deployments, network}) {
     const marketplaceDeployment = await deployments.get("Marketplace")
 
     if (marketplaceDeployment.address === MARKETPLACE_HARDCODED_ADDRESS) {
-      console.log('Marketplace is deployed to expected address. Aborting aliasing.')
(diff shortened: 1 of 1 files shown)


deploy/marketplace.js
```diff
@@ -24,9 +26,24 @@ async function mine256blocks({ network, ethers }) {
   }
 }
 
+async function aliasContract({deployments, network}) {
+  if (network.tags.local) {
+    const marketplaceDeployment = await deployments.get("Marketplace")
+
+    if (marketplaceDeployment.address === MARKETPLACE_HARDCODED_ADDRESS) {
+      console.log('Marketplace is deployed to expected address. Aborting aliasing.')
```
emizzle, line 34: Probably don't need this, as it is expected behaviour.

emizzle (approved):
Nice work Adam, thanks for doing this.

PR #61, opened 2023-06-21
Support slot queue

emizzle:
1. add request expiry to StorageRequested evcent
2. add slotIndex to SlotFreed event
3. remove slotIndex as an indexed emission of SlotFIlled event
4. copy slotIndex before deleting slot, as it was emitting 0!

### DIFF
--- a/contracts/Marketplace.sol
+++ b/contracts/Marketplace.sol
@@ -170,7 +170,7 @@ contract Marketplace is Proofs, StateRetrieval {
 
     delete _slots[slotId];
     context.slotsFilled -= 1;
-    emit SlotFreed(requestId, slotId);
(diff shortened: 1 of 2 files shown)


AuHau (approved):
LGTM 👍

contracts/Marketplace.sol
```diff
@@ -168,9 +168,10 @@ contract Marketplace is Proofs, StateRetrieval {
 
     _removeFromMySlots(slot.host, slotId);
 
+    uint256 slotIndex = slot.slotIndex;
```
AuHau, line 171: Interesting! 😅

emizzle: I'm not sure how the nim-codex contracts tests were passing without this 🤔 Perhaps nim-codex was on an older commit without this change 🤷

PR #65, opened 2023-08-09
Ethers v6 support

AuHau:
This is blocked until `hardhat-waffle` gets updates its dependencies: https://github.com/NomicFoundation/hardhat-waffle/issues/26


2-towns:
@AuHau Can we close this in favor of #231 ?

PR #68, opened 2023-09-13
Document use of `increaseAllowance`

emizzle:
Documents the use of `increaseAllowance` in OpenZeppelin's `ERC20` implementation. 

The `increaseAllowance` function of OpenZeppelin's ERC20 implementation is
being used, which is not part of the standard ERC20 interface. If the Codex
token must be an ERC20 token (eg DAI), then that token will need to be wrapped in
OpenZeppelin's ERC20Wrapped to ensure that `increaseAllowance` is available.

The reason `increaseAllowance` is needed is because the sales state machine
handles concurrent slot fills, which requires token approval. Because these
approvals happen concurrently, simply calling `approve` is not sufficient as
each `approve` amount will overwrite the previous. As an example, if a single
node attempts to fill two slots simultaneously, it will send two `approve`
transactions followed by two `fillSlot` transactions. The first two `approve`
transactions will each approve the collateral amount for one `fillSlot`, the
second overwriting the first, so the total approved will only ever be enough for
one `fillSlot`. So after the first `fillSlot` is sent, the second one will fail
as not enough tokens will have been approved. The solution is use
`increaseAllowance` instead of `approve` which increases the allowance instead
of overwriting it.

### DIFF
--- a/contracts/TestToken.sol
+++ b/contracts/TestToken.sol
@@ -3,6 +3,24 @@ pragma solidity ^0.8.0;
 
 import "@openzeppelin/contracts/token/ERC20/ERC20.sol";
 
+/* The `increaseAllowance` funcation of OpenZeppelin's ERC20 implementation is
+being used, which is not part of the standard ERC20 interface. If the Codex
+token must be an ERC20 token (eg DAI), then that token will need to be wrapped in
+OpenZeppelin's ERC20Wrapped to ensure that `increaseAllowance` is available.
+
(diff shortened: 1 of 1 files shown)


emizzle:
May no longer be needed: https://github.com/codex-storage/nim-codex/issues/550#issuecomment-1754481085

markspanbroek:
Closing. Will be handled in https://github.com/codex-storage/nim-codex/issues/598

PR #72, opened 2023-10-18
Remove PoR parameters, add merkle root for storage proofs

markspanbroek:
Update Request to accomodate new proving scheme:
- Remove PoR parameters
- Remove total number of chunks
- Add merkle root of the dataset

### DIFF
--- a/contracts/Requests.sol
+++ b/contracts/Requests.sol
@@ -23,19 +23,8 @@ struct Ask {
 }
 
 struct Content {
-  string cid; // content id (if part of a larger set, the chunk cid)
-  Erasure erasure; // Erasure coding attributes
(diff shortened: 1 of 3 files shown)


markspanbroek:
Rationale for removing "total number of chunks" from the request:
1. We don't need it to download the slot data. The cid points to the manifest, and the manifest provides enough information to determine which range of blocks to download for a given slot index.
2. We don't need it for the SNARK proofs, because it only needs to know how many shards there are in a slot, and it can calculate that from the slot size (already on chain) and the shard size (fixed in the SNARK)

AuHau (approved):
👍

AuHau:
Is this related to https://github.com/codex-storage/nim-codex/issues/579 ?

PR #74, opened 2023-11-03
use SafeERC20

0xb337r007:
requires #73 to be merged before this one.

This PR add the use of `safeTransferFrom` from `SafeERC20` to avoid checking the return values on transfer.


markspanbroek (approved):
Perfect, thanks!

markspanbroek:
Converted to draft PR, to ensure that it doesn't get merged before #73 is.

0x-r4bbit:
@markspanbroek @emizzle I believe we can close this PR until we know exactly what token the protocol is going to work with.

PR #79, opened 2024-01-05
feat: verifier integration

AuHau:
This is a draft PR of my integration of the verifier, which I managed to get to before going to vacation. I managed to get to the basic integration point, but further work is needed to finish it.

The current state is that the Verifier is deployed and works with the generated proof that is part of the Circuit assets build [here](https://github.com/codex-storage/codex-storage-proofs-circuits/actions/workflows/generate.yml). The tests `Proofs.test.js` are passing (which was my personal goal), but the whole suite not yet.

There are numerous TODOs spread around the changes to better indicate what needs to be finished, but the main high-level things are:
- [x] Network dependant Verifier deployment - have different verifiers for different networks
- [x] Figure out how to assemble the public signals as those should be known to the contract and hence does not need to be passed into the `submitProof()` call
- [x] "Make it work and polished" 😅


test/Proofs.test.js
```diff
@@ -192,25 +195,26 @@ describe("Proofs", function () {
     })
 
     it("submits a correct proof", async function () {
-      await proofs.submitProof(slotId, proof)
+      await proofs.submitProof(slotId, ...proof)
     })
 
     it("fails proof submission when proof is incorrect", async function () {
-      await expect(proofs.submitProof(slotId, [])).to.be.revertedWith(
+      await expect(proofs.submitProof(slotId, ["0x1bcdb9a3c52070f56e8d59b29239f0528817f99745157ce4d03faefddfff6acc", "0x2496ab7dd8f0596c21653105e4af7e48eb5395ea45e0c876d7db4dd31b4df23e"],[["0x002ef03c350ccfbf234bfde498378709edea3a506383d492b58c4c35ffecc508", "0x174d475745707d35989001e9216201bdb828130b0e78dbf772c4795fa845b5eb"],["0x1f04519f202fac14311c65d827f65f787dbe01985044278292723b9ee77ce5ee", "0x1c42f4d640e94c28401392031e74426ae68145f4f520cd576ca5e5b9af97c0bb"]],["0x1db1e61b32db677f3927ec117569e068f62747986e4ac7f54db8f2acd17e4abc", "0x20a59e1daca2ab80199c5bca2c5a7d6de6348bd795a0dd999752cc462d851128"],["0x00000000000000000000000000000000000000000000000000000001b9b78422","0x2389b3770d31a09a71cda2cb2114c203172eac63b61f76cb9f81db7adbe8fc9d","0x0000000000000000000000000000000000000000000000000000000000000003"]))
```
emizzle, line 202: Just a general observation: it's difficult to know which part of this input makes the proof invalid without comparing all the values to the values in `proofs.json` and `public.json` (or returned from `loadProof`). A guess/suggestion to make it more clear: provide an incorrect slot index.

emizzle (commented):
Looks great Adam! This is very much inline with what I had in mind when we first discussed it in the marketplace meeting.

markspanbroek:
Should be good for another review now @AuHau & @emizzle.

contracts/Endian.sol
```diff
@@ -0,0 +1,14 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.23;
+
+contract Endian {
+  /// reverses byte order to allow conversion between little endian and big
+  /// endian integers
+  function _byteSwap(bytes32 input) internal pure returns (bytes32 output) {
```
dryajov, line 7: Wouldn't it be much faster to do the conversion outside the contract?

AuHau: It would (and be cheaper as well), but the data that are converted with this are gathered from the contract's data and are not "input to function call", so IMHO not really possible without removing the goal that we don't want to submit the public inputs to the `submitProof()` call. Have a look here: https://github.com/codex-storage/codex-contracts-eth/pull/79/files#diff-5973808e32384782b5978a63e85b893b67d634be8aa42001f0ef480f189a0688R162-R193

AuHau: But thinking about it a bit more, maybe this conversion would in itself warrant actually dropping that goal in order to optimize it from a gas perspective. I don't know what is cheaper though. IMHO we don't need to analyze it now, but it would be good to have a look on it later on with some analysis.

AuHau (commented):
LGTM! Pretty good work. Thanks for taking it over and finishing it! 👍 I can't approve my own PR, so somebody else needs to do that. 🤪

Just a side note, there were a lot of formatting changes. It is kind of distracting for reviewing... Maybe in the future, I would adapt some aggressive styler like Prettier just for Solidity to prevent these in the future.

contracts/Endian.sol
markspanbroek: We cannot allow hosts to supply their own public inputs, that would defeat the security of the scheme. They have to come from values that are on-chain. 

What we could do for the two values that are converted is the following:
- for the merkle root in the storage request, we can change its definition from a bytes32 to a uint256, which gets rid of the need to interpret bytes as an integer (be it little or big endian)
- for the randomness, we could interpret that as a big-endian integer, instead of a little-endian integer in the nim code

This however has two disadvantages:
- when we move to sha256 instead of poseidon2, we have to change the type of the merkle root back to bytes32 again
- we get a scheme for the zero-knowledge proof where most of the time we interpret bytes as little-endian, but in a special case we interpret bytes as big-endian

AuHau: Yeah, I did not really though it through. I thought that "we would do assertions on the submitted public inputs" to verify if it is not "defeating the security", but that would only circle back to the need to the conversions 😅 Never mind.

dryajov: There isn't any particular reason we're using little endian in the client side, why not change it there?

dryajov: In fact, it was probably an oversight that we didn't set off with bigendian from the get go, given that the EVM uses bigendian.

markspanbroek: We chose little endian because the circom and arkworks tooling uses it.

emizzle (approved):
LGTM. Some of this I haven't looked to in depth and don't have any smart ideas around the endian conversion. However, if we do come up with something, we can submit a new PR for it.

markspanbroek:
I'm merging this, we can always update the code when we think of a better way to handle endianness.

PR #82, opened 2024-01-25
Dummy verifier

markspanbroek:
Deploy 2 versions of the marketplace (only on local network):
- Marketplace with Groth16 Verifier
- Marketplace with TestVerifier

The test verifier accepts all proofs, except when it has all 0 values. 
Necessary for testing the codex client, until we can generate real ZK proofs.

Removes the code that deploys contracts to a fixed address, because it doesn't seem to work, and I couldn't find a way to get it to work.


AuHau (approved):
LGTM

PR #83, opened 2024-01-30
Refactor of the solidity Groth16 verifier

markspanbroek:
Refactors the Groth16 verifier to:
- fix mistakes
- remove unnecessary conversions
- make it more readable
- return false instead of stopping execution when an operation fails

### DIFF
--- a/contracts/Groth16Verifier.sol
+++ b/contracts/Groth16Verifier.sol
@@ -131,8 +131,11 @@ contract Groth16Verifier is IGroth16Verifier {
     Groth16Proof calldata proof,
     uint[] memory input
   ) public view returns (bool success) {
-    require(input.length + 1 == _verifyingKey.ic.length, "verifier-bad-input");
-    // Check that inputs are field elements
+    // Check amount of public inputs
+    if (input.length + 1 != _verifyingKey.ic.length) {
+      return false;
+    }
+    // Check that public inputs are field elements
     for (uint i = 0; i < input.length; i++) {
       if (input[i] >= _Q) {
         return false;
@@ -151,6 +154,7 @@ contract Groth16Verifier is IGroth16Verifier {
         return false;
       }
     }
+    // Check the pairing
     uint outcome;
     (success, outcome) = _checkPairing(
       _negate(proof.a),
--- a/contracts/Groth16Verifier.sol
+++ b/contracts/Groth16Verifier.sol
@@ -160,12 +160,18 @@ contract Groth16Verifier {
       );
       G1Point memory product;
       (success, product) = Pairing.multiply(_verifyingKey.ic[i + 1], input[i]);
-      require(success, "pairing-mul-failed");
+      if (!success) {
+        return false;
+      }
       (success, vkX) = Pairing.add(vkX, product);
-      require(success, "pairing-add-failed");
+      if (!success) {
+        return false;
+      }
     }
     (success, vkX) = Pairing.add(vkX, _verifyingKey.ic[0]);
-    require(success, "pairing-add-failed");
+    if (!success) {
+      return false;
+    }
     uint outcome;
     (success, outcome) =
       Pairing.pairingProd4(
@@ -178,7 +184,9 @@ contract Groth16Verifier {
         proof.c,
         _verifyingKey.delta2
       );
-    require(success, "pairing-opcode-failed");
+    if (!success) {
+      return false;
+    }
     return outcome == 1;
   }
 }
--- a/contracts/Groth16Verifier.sol
+++ b/contracts/Groth16Verifier.sol
@@ -122,14 +122,7 @@ contract Groth16Verifier {
 
     // solhint-disable-next-line no-inline-assembly
     assembly {
-      success := staticcall(
-        sub(gas(), 2000),
-        8,
-        input,
-        768, // 24 uints, 32 bytes each
-        output,
-        32
-      )
+      success := staticcall(sub(gas(), 2000), 8, input, 768, output, 32)
     }
     return (success, output[0]);
   }
--- a/contracts/Groth16Verifier.sol
+++ b/contracts/Groth16Verifier.sol
@@ -27,8 +27,7 @@ library Pairing {
 
   /// The negation of p, i.e. p.addition(p.negate()) should be zero.
   function negate(G1Point memory p) internal pure returns (G1Point memory) {
-    if (p.x == 0 && p.y == 0) return G1Point(0, 0);
-    return G1Point(p.x, _Q - (p.y % _Q));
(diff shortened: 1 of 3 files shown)


contracts/Groth16Verifier.sol
```diff
@@ -20,107 +20,66 @@
 pragma solidity 0.8.23;
 import "./Groth16.sol";
 
-library Pairing {
-  // The prime q in the base field F_q for G1
-  uint private constant _Q =
+contract Groth16Verifier is IGroth16Verifier {
+  uint private constant _P =
     21888242871839275222246405745257275088696311157297823662689037894645226208583;
+  uint256 private constant _Q =
```
bkomuves, line 26: This one should definitely be called `R` instead of `Q`. Using `r` for the size of the scalar field (the one ending with ...17) is essentially an universal standard. For the other one (here `P`) some people use `p` and some people use `q`).

AuHau (commented):
Honestly I have very limited understanding of what is going on here 😅 Maybe @bkomuves could do proper review? If not then I will dive in, but it will take me quite some time to understand it what is going on and what has changed.

markspanbroek:
> Honestly I have very limited understanding of what is going on here 😅 Maybe @bkomuves could do proper review? If not then I will dive in, but it will take me quite some time to understand it what is going on and what has changed.

That's ok, nobody in our team knows everything about zero knowledge proofs and solidty. That's why I asked multiple people to review. Balazs can review the crypto bits, and you and Eric can look for solidity mistakes.

Things that you could focus a review on:
- were the refactor steps done properly (each one is its own commit, so they can be viewed separately)
- looking at the end result, does anything stand out w.r.t solidty types/visibility/etc..
- are the assembly calls to the precompiled contracts done correctly, and do they specify the correct number of input and output bytes

For reference, this is what I used to look up the precompiled contracts:
https://www.evm.codes/precompiled

And this is what I used as a reference for the STATICCALL opcode:
https://www.evm.codes/#fa?fork=shanghai

contracts/Groth16Verifier.sol
```diff
@@ -20,107 +20,66 @@
 pragma solidity 0.8.23;
 import "./Groth16.sol";
 
-library Pairing {
-  // The prime q in the base field F_q for G1
-  uint private constant _Q =
+contract Groth16Verifier is IGroth16Verifier {
+  uint private constant _P =
```
bkomuves, line 24: `_P` is `uint` but `_Q` is `uint256`. I'm not sure if this makes any difference, but should be consistent anyway? (personally I would use `uint256` everywhere because that way it's more explicit that these really big numbers; also in `Groth16.sol`)

markspanbroek: Yes, good catch. They are the same, but `uint256` is indeed preferred.

contracts/Groth16Verifier.sol
```diff
@@ -20,107 +20,66 @@
...
-  ) internal view returns (G1Point memory r) {
-    uint[4] memory input;
-    input[0] = p1.x;
-    input[1] = p1.y;
-    input[2] = p2.x;
-    input[3] = p2.y;
-    bool success;
-    // solhint-disable-next-line no-inline-assembly
-    assembly {
-      success := staticcall(sub(gas(), 2000), 6, input, 0xc0, r, 0x60)
```
bkomuves, line 47: These `sub(gas(),2000)` things look very strange. I guess that would mean to pass the "available amount minus 2000" gas to the precompile? I guess technically it should work, but why 2000? or why subtract anything at all?

According to the docs I found, the first parameter of `staticcall` is

>  gas: amount of gas to send to the sub context to execute. The gas that is not used by the sub context is returned to this one.

Based on this just `gas()` would work? Also, what even does `gas()` mean, I cannot find it in the documentation... Is it maybe a deprecated name for `gasleft()`?

bkomuves: Btw, the pairing precompile documentation says that 

"If the input is not valid, all gas provided is consumed and there is no return data, indicating a precompile contract error."

so maybe you shouldn't actually pass all the gas available, because that makes failures more expensive?
(not that there should be failures in normal working, why would anybody spend gas by sending an incorrect proof...)

markspanbroek: Good question, I left this unchanged from the original code, but I don't know why they subtract 2000 either.

The `gas()` call is this EVM opcode: https://www.evm.codes/#5a?fork=shanghai
It should indeed return gas left.

bkomuves (commented):
Apart from the minor issues mentioned in comments, the verifier contract looks good to me. 

I would add a comment explaining that the 3 precompiles are all supposed to check for the validity of the group elements, with a link to a definitive source.

I can also imagine that with many different EVM implementations, some may forget to actually do the checks, but I guess that's not our problem? :)

We could also add our own group checks to be extra safe, the gas cost would be probably a few hundred (small compared to the whole verification, at least on Ethereum mainnet), but not sure if it's worth the complexity?

contracts/Groth16Verifier.sol
bkomuves: Oh I was looking at the wrong place, this is assembly opcode and I was looking at the solidity builtins...

Still the minus 2000 doesn't make much sense to me, but I'm very much a non-expert in EVM :)

markspanbroek: I looked into this a bit further, and decided to just go with `gas()` as you suggested.

Even though EIP-196 states that it "consumes all gas provided" when the inputs are invalid, I cannot reproduce that on hardhat, and looking through the Geth code, it also doesn't seem to do that (https://github.com/ethereum/go-ethereum/blob/b80643b7370075262fd6dfad7ae8aa77710e2ef1/core/vm/contracts.go#L484). 
Also, strangely, the accompanying EIP-197 does not state that it consumes all gas, even though it is a similar use case.

And in the end it doesn't matter whether or not it consumes all the gas. If it does, then verification fails with an 'out-of-gas' error, and if it doesn't, then `verify` returns false. In both cases the proof is not verified.

markspanbroek:
> I would add a comment explaining that the 3 precompiles are all supposed to check for the validity of the group elements, with a link to a definitive source.

Good idea, I added the comments.

> I can also imagine that with many different EVM implementations, some may forget to actually do the checks, but I guess that's not our problem? :)

That would lead to a consensus failure, and seems unlikely. The EIPs state that these checks should be done, and looking at the Geth code, it seems to do these checks.

markspanbroek:
I think that this is ready to be merged. @AuHau or @emizzle, could you do a final review to check for any Solidity shenanigans? Then I can merge this :pray:

AuHau (approved):
Had a look and could not find anything questionable, so it is LGTM from me. Nice work! 👍

PR #84, opened 2024-01-31
Fix deployment bug in loading zkey hash

markspanbroek:
Fixes bug introduced in #82


AuHau (approved):
Ups, thanks for the catch!

PR #85, opened 2024-02-06
Explicit getters for token and config

markspanbroek:
Implicit getters have slightly different semantics when it comes to ABI encoding their results.
This lead to a problem when calling the getter for the config that contains the zkeyHash string.

The decoding problem has been solved in https://github.com/codex-storage/nim-ethers/pull/63, but that is not ready to be merged yet, because it depends on the chronos v4 changes in nim-ethers.

This is a workaround that makes the getters explicit.


AuHau (approved):
LGTM 👍

emizzle (commented):
LGTM too 👍

PR #87, opened 2024-02-20
Verifier: represent elements from field F_{p^2} as `real + i * imag`

markspanbroek:
Reason: Circom and Ethereum EIP-197 both represent these elements as arrays of two elements, but they do it in reverse order of each other. This change makes it explicit which number is the real part, and which number is the imaginary part.

### DIFF
--- a/contracts/Groth16.sol
+++ b/contracts/Groth16.sol
@@ -6,9 +6,17 @@ struct G1Point {
   uint y;
 }
 
+// A field element F_{p^2} encoded as `real + i * imag`.
+// We chose to not represent this as an array of 2 numbers, because both Circom
+// and Ethereum EIP-197 encode to an array, but with conflicting encodings.
+struct Fp2Element {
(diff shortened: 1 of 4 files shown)


contracts/Groth16.sol
```diff
@@ -6,9 +6,17 @@ struct G1Point {
   uint y;
 }
 
+// A field element F_{p^2} encoded as `real + i * imag`.
+// We chose to not represent this as an array of 2 numbers, because both Circom
+// and Ethereum EIP-197 encode to an array, but with conflicting encodings.
+struct Fp2Element {
+  uint real;
```
bkomuves, line 13: Just a minor thing: Earlier in another PR we had a discussion about that using `uint256` for these numbers maybe better conveys the intent of these really being big numbers.

bkomuves (approved):
Looks good to me (it's a very straightforward change)

PR #88, opened 2024-02-20
Increase proof period

markspanbroek:
In Codex's integration tests we now create real ZK proofs, which take a bit longer to generate. We therefore need a period that remains the same while the proof is generated.

### DIFF
--- a/deploy/marketplace.js
+++ b/deploy/marketplace.js
@@ -9,8 +9,10 @@ const CONFIGURATION = {
     slashPercentage: 10,
   },
   proofs: {
-    period: 10,
-    timeout: 5,
+    period: 60,
+    timeout: 30,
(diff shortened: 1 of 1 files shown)


dryajov (approved):

emizzle:
I'm not sure that increasing the `timeout` will help with an increased proof generation time, as the validator is checking the previous period, so the `timeout` actually reflects how long the validator has to mark a proof as missing for the *previous* period.

If the validator were to generate proofs, then I do believe this should be increased.

markspanbroek:
> I'm not sure that increasing the `timeout` will help with an increased proof generation time

You're right, it doesn't help with that :smile:  I just changed it to keep the timeout at the halfway mark of the period.

PR #89, opened 2024-02-23
chore: add verifier assets for codexdisttestnetwork

AuHau:
~~This temporary fixes the non-functional distributed tests, which were failing on the non-existing `PUSH0` opcode. This opcode was introduced in the Shangai upgrade.~~

~~I originally tried to upgrade the Geth setup in dist. tests to support the Shangai upgrade, but with having `--miner` functionality Geth refused to enable it. It turned out that as it is Post-Merge upgrade it would require more complicated setup (https://github.com/ethereum/go-ethereum/issues/28449 https://dev.to/q9/how-to-merge-an-ethereum-network-right-from-the-genesis-block-3454).~~

~~So instead I decided to lower the Solidity's compile EVM target to pre-Shangai upgrade as a temporary measure as first version of the testnet will run against our PoA network. When we will deploy to some existing testnet with Shangai support we should revisit this, although then we will also need a solution for dist tests. Maybe the Geth's `--dev` mode could be then the answer? Not sure.~~

EDIT: Ben was able to create Geth node image with Shangai upgrade, so the reverting part of the compiled target is not necessary. Only thing left for having the contracts successfully deployed on Dist. test is the missing verifier assets for that network.  I have  added them and they are for now identical to `hardhat` network.


markspanbroek (approved):

markspanbroek:
Thanks for looking into this Adam!

PR #92, opened 2024-03-04
from the codex-storage-proofs-circuits commit 066bd73

benbierens:


markspanbroek:
Were the r1cs and wasm files unchanged? Otherwise we'll need updated versions of these too.

benbierens:
> Were the r1cs and wasm files unchanged? Otherwise we'll need updated versions of these too.

I did pull in the matching r1cs and wasm. They were unchanged according to git. 👍

PR #94, opened 2024-03-15
Update .tool-versions

elcritch:
I use ASDF and it kept trying to use this older node version.


markspanbroek (approved):
Nice, another ASDF user :smile:

elcritch:
> Nice, another ASDF user 😄

Yah for a long time!

PR #97, opened 2024-03-27
chore(tests): move circuit artifacts

emizzle:
All artifacts, apart from the `.circom` artifact, have been moved to `nim-codex/tests/circuits/fixtures`

`proof_main.circom` was removed, as it was used for reference only.


dryajov:
We should probably leave this as is, it's OK for the contracts to use their own versions for unit testing/integration. The only thing I wanted, is for codex to use local artifacts instead of depending on a third party package, but here it should be fine to keep this as is.

markspanbroek:
Looks like this is no longer needed, closing.

PR #99, opened 2024-04-22
feat: expiry specified with number of seconds

AuHau:
This is part of the work on fixing https://github.com/codex-storage/nim-codex/issues/766.

Not completely sure why the `expiry` was designed in this way where a timestamp was passed, while `duration` is computed as `now + duration interval`. Originally for the fix mentioned above, I thought I would do this computation in `nim-codex` but I ran into a clock syncing issue in tests, so I have instead decided to do a proper fix with unifying the "time computation" in the smart contracts for expiry and duration in the same way.

### DIFF
--- a/contracts/Marketplace.sol
+++ b/contracts/Marketplace.sol
@@ -32,6 +32,7 @@ contract Marketplace is Proofs, StateRetrieval, Endian {
     uint256 expiryFundsWithdraw;
     uint256 startedAt;
     uint256 endsAt;
+    uint256 expiresAt;
   }
 
   struct Slot {
@@ -90,11 +91,14 @@ contract Marketplace is Proofs, StateRetrieval, Endian {
 
     RequestId id = request.id();
     require(_requests[id].client == address(0), "Request already exists");
+    require(
+      request.expiry > 0 && request.expiry < request.ask.duration,
+      "Expiry not between 0 and duration"
+    );
 
     _requests[id] = request;
-    uint256 endsAt = block.timestamp + request.ask.duration;
-    require(endsAt > request.expiry, "Request end before expiry");
-    _requestContexts[id].endsAt = endsAt;
+    _requestContexts[id].endsAt = block.timestamp + request.ask.duration;
+    _requestContexts[id].expiresAt = block.timestamp + request.expiry;
 
     _addToMyRequests(request.client, id);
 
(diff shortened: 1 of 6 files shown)


contracts/Marketplace.sol
```diff
@@ -103,7 +107,7 @@ contract Marketplace is Proofs, StateRetrieval, Endian {
     _marketplaceTotals.received += amount;
     _transferFrom(msg.sender, amount);
 
-    emit StorageRequested(id, request.ask, request.expiry);
+    emit StorageRequested(id, request.ask, _requestContexts[id].expiresAt);
```
emizzle, line 110: I think this is definitely the way to go. The node will need to consume this value and set the local timer accordingly instead of calculating the value. The pre-calculation of `expiry` on the node end was possibly why a timestamp was initially used for expiry. 

We could probably update the `purchasing/states/pending` state to listen for the `StorageRequested` event before going to the `submitted` state, passing in the `expiresAt` value from the event to the `submitted` state. In the `submitted` state, we can use `await clock.waitUntil(expiresAt)` and cancel that future if the request is fulfilled.

contracts/Marketplace.sol
```diff
@@ -90,11 +91,14 @@ contract Marketplace is Proofs, StateRetrieval, Endian {
 
     RequestId id = request.id();
     require(_requests[id].client == address(0), "Request already exists");
+    require(
+      request.expiry > 0 && request.expiry < request.ask.duration,
+      "Expiry not between 0 and duration"
```
emizzle, line 96:
```suggestion
      "Expiry not between 0 and duration (exclusive)"
```

test/Marketplace.test.js
```diff
@@ -478,7 +489,7 @@ describe("Marketplace", function () {
 
       const endBalance = await token.balanceOf(host.address)
       const expectedPartialPayout =
-        (request.expiry - fillTimestamp) * request.ask.reward
+        (request.expiry - fillSeconds - 1) * request.ask.reward // TODO: I don't understand why there needs to be -1 offset
```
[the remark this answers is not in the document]
AuHau: Found it!
You are right about some mined transaction that advances the clock but it is not of `freeSlot` call. It does not matter at what block's timestamp you are calling the `freeSlot` as the payout is always for the interval `<filledAt;expiryAt>`.

It was because of the `approve()` call in the `beforeEach` statement that advanced the clock after `requestStorage`. TBH the way how I am calculating the partial payout is not really that "fool-proof" 😅

PR #118, opened 2024-06-12
docs: proofs comments

AuHau:
As part of work on defining Application Properties for Certora you can find here: https://notes.status.im/s/D29-EqKzi

I have been writing comments in the code to explain the code's functionality better.

### DIFF
--- a/contracts/Proofs.sol
+++ b/contracts/Proofs.sol
@@ -140,7 +140,7 @@ abstract contract Proofs is Periods {
   }
 
   /**
-   * @see isProofRequired
+   * See isProofRequired
    */
   function _isProofRequired(
     SlotId id,
--- a/contracts/Proofs.sol
+++ b/contracts/Proofs.sol
@@ -6,10 +6,18 @@ import "./Requests.sol";
 import "./Periods.sol";
 import "./Groth16.sol";
 
+/**
+ * @title Proofs
+ * @notice Abstract contract that handles proofs tracking, validation and reporting functionality
+ */
 abstract contract Proofs is Periods {
   ProofConfig private _config;
   IGroth16Verifier private _verifier;
 
+  /**
+   * Creation of the contract requires at least 256 mined blocks!
+   * @param config Proving configuration
+   */
   constructor(
     ProofConfig memory config,
     IGroth16Verifier verifier
@@ -19,27 +27,46 @@ abstract contract Proofs is Periods {
     _verifier = verifier;
   }
 
-  mapping(SlotId => uint256) private _slotStarts;
+  mapping(SlotId => uint256) private _slotStarts; // TODO: Should be smaller then uint256
   mapping(SlotId => uint256) private _probabilities;
-  mapping(SlotId => uint256) private _missed;
+  mapping(SlotId => uint256) private _missed; // TODO: Should be smaller then uint256
   mapping(SlotId => mapping(Period => bool)) private _received;
   mapping(SlotId => mapping(Period => bool)) private _missing;
 
   function slotState(SlotId id) public view virtual returns (SlotState);
 
+  /**
+   * @return Number of missed proofs since Slot was Filled
+   */
   function missingProofs(SlotId slotId) public view returns (uint256) {
     return _missed[slotId];
   }
 
+  /**
+   * @param slotId Slot's ID for which the proofs should be resetted
+   * @notice Resets the missing proofs counter to zero
+   */
   function _resetMissingProofs(SlotId slotId) internal {
     _missed[slotId] = 0;
   }
 
+  /**
+   * @param id Slot's ID for which the proofs should be started to require
+   * @param probability Integer which specifies the probability of how often the proofs will be required. Lower number means higher probability.
+   * @notice Notes down the block's timestamp as Slot's starting time for requiring proofs
+   *     and saves the required probability.
+   */
   function _startRequiringProofs(SlotId id, uint256 probability) internal {
     _slotStarts[id] = block.timestamp;
     _probabilities[id] = probability;
   }
 
+  /**
+   * @param id Slot's ID for which the pointer should be calculated
+   * @param period Period for which the pointer should be calculated
+   * @return Uint8 pointer that is stable over current Period and points to one block over the last 256 block's window.
+   * @dev For more information see https://github.com/codex-storage/codex-research/blob/master/design/storage-proof-timing.md
+   */
   function _getPointer(SlotId id, Period period) internal view returns (uint8) {
     uint256 blockNumber = block.number % 256;
     // To ensure the pointer does not remain in downtime for many consecutive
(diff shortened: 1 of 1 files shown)


contracts/Proofs.sol
```diff
@@ -19,27 +27,46 @@ abstract contract Proofs is Periods {
...
+   */
   function _startRequiringProofs(SlotId id, uint256 probability) internal {
     _slotStarts[id] = block.timestamp;
     _probabilities[id] = probability;
   }
 
+  /**
+   * @param id Slot's ID for which the pointer should be calculated
+   * @param period Period for which the pointer should be calculated
+   * @return Uint8 pointer that is stable over current Period and points to one block over the last 256 block's window.
```
emizzle, line 67:
```suggestion
   * @return Uint8 pointer that is stable over current Period, ie points to same block in the Period window of 256 blocks when new blocks are produced.
```

contracts/Proofs.sol
```diff
@@ -51,27 +78,49 @@ abstract contract Proofs is Periods {
     return uint8(pointer);
   }
 
+  /**
+   * @param id Slot's ID for which the pointer should be calculated
+   * @return Uint8 pointer that is stable over current Period and points to one block over the last 256 block's window.
```
emizzle, line 83: Same as above

contracts/Proofs.sol
```diff
@@ -51,27 +78,49 @@ abstract contract Proofs is Periods {
...
+   * @dev For more information see https://github.com/codex-storage/codex-research/blob/master/design/storage-proof-timing.md
+   */
   function getPointer(SlotId id) public view returns (uint8) {
     return _getPointer(id, _blockPeriod());
   }
 
+  /**
+   * @param pointer Pointer that points to one block in the last 256 blocks window
```
emizzle, line 91:
```suggestion
   * @param pointer Integer [0-255] that indicates an offset of the last 256 blocks, pointing to a block that remains constant for the entire Period's duration.
```

contracts/Proofs.sol
AuHau:
```suggestion
   * @return Uint8 pointer that is stable over current Period, ie an integer offset [0-255] of the last 256 blocks, pointing to a block that remains constant for the entire Period's duration.
```

contracts/Proofs.sol
```diff
@@ -19,27 +27,46 @@ abstract contract Proofs is Periods {
...
+   * @param slotId Slot's ID for which the proofs should be resetted
+   * @notice Resets the missing proofs counter to zero
+   */
   function _resetMissingProofs(SlotId slotId) internal {
     _missed[slotId] = 0;
   }
 
+  /**
+   * @param id Slot's ID for which the proofs should be started to require
+   * @param probability Integer which specifies the probability of how often the proofs will be required. Lower number means higher probability.
```
emizzle, line 55: I actually don’t know how to describe `proofProbability`. Only when the challenge 
 modulo the proofProbability is 0 is there a proof required. And since the challenge is random, it seems like the proofProbability doesn’t change the probability of a proof being required.

emizzle (approved):
Might need to get @markspanbroek’s input on it, but don’t want to hold up the work while he’s away.

AuHau:
Oh right, that makes sense. It is not crucial, I just wanted to merge it so I don't have to think about it, but yeah let's wait for Mark's input.

contracts/Proofs.sol
gmega: A higher number in the modulo means there's a smaller percentage of the total set of possible entropy numbers that will result in zero. Think of the degenerate case of when you use a number in the modulo that's larger than half of the maximum number allowed for the entropy: you'll get exactly one number for which this is zero. Since you're drawing numbers at random, the chance you hit this is $1/n$, where $n$ is the maximum value for the entropy.

If you use something like $2$, on the other hand, half of the random numbers you get will result in zero, so you get a proof probability of $50\%$.

AuHau: Yeah, honestly that is something I am not such a fan of, because you can not pretty much specify any probability between 100% and 50% if I understand this correctly, right @gmega ?

gmega: Yeah, you're missing a bunch of points, really, cause you're getting $\frac{1}{n}$ where $n$ is a natural number. So you get  $100\\%, 50\\%, 33.34\\%, 25\\%, \cdots$

I believe this is done like that cause it's easier to implement. As to whether or not this is a problem, I guess it depends on the period length. If you have short periods (e.g. block time of Ethereum), then maybe it's not a problem as a setting above $50\\%$ would probably never be used anyway.

gmega: Btw, regardless, I think that description is inaccurate as $\text{probability}$ does not specify a probability, $1/\text{probability}$ does, and even then it's the probability that a proof will be required at a single discrete time interval, which I guess you're calling a `Period` in the code. 

I'd be tempted to say that this describes _the average number of `Period`s between proofs_, but since we do some [funky stuff with the number I don't understand that well](https://github.com/codex-storage/codex-research/blob/master/design/storage-proof-timing.md#pointer-downtime), I'm afraid this would also be misleading. But this is definitely not a probability.

AuHau: > Yeah, you're missing a bunch of points, really, cause you're getting  where  is a natural number. So you get...

That is the thing - it is not natural number but integer (`uint256`): https://github.com/codex-storage/codex-contracts-eth/blob/master/contracts/Requests.sol#L19. Otherwise, I would agree with you...

emizzle: Thanks for the explanation, makes much more sense now @gmega  🙏 

> perhaps more loosely the proof frequency
Yes, I agree it makes much more sense to call this proof frequency, where we'd expect, on average, a proof to be required every $n$ proofs

contracts/Proofs.sol
```diff
@@ -97,17 +152,40 @@ abstract contract Proofs is Periods {
...
     return _isProofRequired(id, _blockPeriod());
   }
 
+  /**
+   * Proof Downtime specifies part of the Period when the proof is not required even
+   * if the proof should be required. This function returns true if the pointer is
+   * in downtime (hence no proof required now) and at the same time the proof
+   * will be required later on in the Period.
+   *
+   * @dev see for more info about downtime: https://github.com/codex-storage/codex-research/blob/master/design/storage-proof-timing.md#pointer-downtime
```
markspanbroek, line 169: Permanent link like above

contracts/Proofs.sol
```diff
@@ -97,17 +152,40 @@ abstract contract Proofs is Periods {
...
+   * @param id SlotId for which the proof requirements should be checked. Its state needs to be Filled.
```
markspanbroek, line 170: > Its state needs to be Filled

Same as above

markspanbroek (approved):
Thanks @AuHau, looks good! Just a few small comments.

contracts/Proofs.sol
```diff
@@ -51,27 +78,49 @@ abstract contract Proofs is Periods {
     return uint8(pointer);
   }
 
+  /**
+   * @param id Slot's ID for which the pointer should be calculated
+   * @return Uint8 pointer that is stable over current Period, ie an integer offset [0-255] of the last 256 blocks, pointing to a block that remains constant for the entire Period's duration.
+   * @dev For more information see https://github.com/codex-storage/codex-research/blob/master/design/storage-proof-timing.md
```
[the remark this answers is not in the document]
AuHau:
```suggestion
   * @dev For more information see [timing of storage proofs](https://github.com/codex-storage/codex-research/blob/41c4b4409d2092d0a5475aca0f28995034e58d14/design/storage-proof-timing.md)
```

contracts/Proofs.sol
AuHau:
```suggestion
   * @dev for more info about downtime see [timing of storage proofs](https://github.com/codex-storage/codex-research/blob/41c4b4409d2092d0a5475aca0f28995034e58d14/design/storage-proof-timing.md#pointer-downtime)
```

contracts/Proofs.sol
```diff
@@ -97,17 +152,40 @@ abstract contract Proofs is Periods {
     return isRequired && pointer >= _config.downtime;
   }
 
+  /**
+   * @param id Slot's ID for which the proof requirements should be checked. Its state needs to be Filled.
```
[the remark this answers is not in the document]
AuHau:
```suggestion
   * @param id Slot's ID for which the proof requirements should be checked. If the Slot's state is other than Filled, `false` is always returned. 
```

contracts/Proofs.sol
AuHau:
```suggestion
   * @param id SlotId for which the proof requirements should be checked. If the Slot's state is other than Filled, `false` is always returned. 
```

contracts/Proofs.sol
AuHau: > Not sure I follow - isn't uint* supposed to represent the set of non-negative integers (i.e., the set of natural numbers)?

Yeah, sorry - had a brain fart 😅 I thought you meant real numbers 🙈 But yeah re-reading your original comment, makes sense that the "upper range" (eq. 100%-50%) won't be really used...

PR #123, opened 2024-07-25
feat: Add validator slot id groups

emizzle:
Related to https://github.com/codex-storage/nim-codex/issues/457 and https://github.com/codex-storage/nim-codex/issues/458.

To cover the entire SlotId (uint256) address space, each validator must validate a portion of the SlotId space. When a slot is filled, the SlotId will be put in to a bucket, based on the value of the SlotId and the number of buckets (validators) configured. Similar to `myRequests` and `mySlots`, a function called `validationSlots` can be used to retrieve the `SlotIds` being validated for a particular bucket (validator index). This facilitates loading actively filled slots in need of validation when a validator starts.

The number of validators specified in the network-level configuration specifies the minimum number of validators to deployed to the network. There can be more than that number of validators on the network. In the Codex client, each validator will opt to validate a bucket of the `SlotId` space, by specifying an index of `[0, validators-1]`. It is important a minimum of 1 validator per bucket is deployed on the network, or the configuration value is set to 1. For example, if the `validators` configuration value was set to 3, there should be a minimum of 3 validators deployed on the network, with each one specifying a different validator bucket to cover. 

These changes do not prevent any 1 validator participant from watching the entire SlotId space to potentially earn more.


AuHau:
So I am a bit surprised by this design. I understand that you are taking a similar approach to how we solve similar issues for other concepts (Slots, Requests), but this seems to me a bit over-engineered solution that is not necessary and will incur additional IMHO unnecessary costs (gas) for users of the network. Also while we go with this solution (Validators) for now, we know we will move away from it in the future. Moreover, I don't really like the staticity of the Validator groups, which are specified on the contract deployed.

Let's discuss it on the call today.

emizzle:
The alternative idea proposed by @AuHau (modified for detail and completeness):

1. [Local] Check repostore for previous state downloads, set block `b` to last downloaded block. If there is no previous block in the repostore,  set block `b` to the start of the block the latest contract was deployed at, or if that doesn't exist, to `0`. There are some unknowns around where to get/persist the latest contract deployment block number, with one possible solution being to hardcode this value along with the contract address in the codebase.
2. [RPC requests] Download past `SlotFilled` events from Block `b` to `latest`.
3. [Local] Store the latest block number in the repo store to prevent redownloading historical state for next node restart 
4. [Local] Check if the filled slot's `SlotId` is in the range of the validator. Validator ranges, or buckets, will have to be hardcoded in the codebase, and each validator will elect an index via a cli parameter.
6. [RPC requests] For all slots in range, check the state of each slot. 
8. [Local] If `Filled`: 1) store the filled slot ids in the repo store to prevent redownloading, 2) start validating
9. [Local + RPC] Similar to the slot queue, subscribe to all contract events to maintain the current slot state in the repo store, along with the latest block number.

The potential downsides of this approach are:
1. Imo, this is more of an overengineered solution than what is proposed in the PR for something that is going to be replaced by aggregators. Agreed, from a gas perspective, this is the better solution, however from a perspective of getting a solution in place so that we can have validators for a working solution, I'd rather make the sacrifice of not having to go through the more involved development process of the above.
2. The amount of RPC requests that need to be performed before the validator can start performing its duties on start up could be potentially very long. As an example, each slot state request incurred a response time of 350ms locally when testing on the current devnet, however that is mostly due to the physical location of the request (australia) and the RPC server (europe). But extrapolate that out to potentially thousands of slots (or more).
3. Leaning on our experience with the slot queue, maintaining the slot state locally, while being just an engineering challenge, is not particularly super simple. Meaning it will likely take more time than is obvious to develop.

Ultimately, as @AuHau stated, this is a temporary solution and will most likely be replaced by aggregators in the future. By my estimate, the amount of work involved above is much more than the solution that is currently proposed in this PR. Because of this, in my opinion, I think we should keep what has been already proposed in this PR.

emizzle:
I put some more thought into this and have attempting to map the pros and cons for each design.
# Loading Validator slots on startup


|                 | Local                                                                                                                                                                                                                                                                                                                                                                                                                                      | On chain                                                                                                                                                                                                                                                                                                                                                                                    |
| --------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| Slot buckets    | 🟡 less chance of validation delays<sup>1,2</sup><br/>🟢 no additional gas requirements<br>🔴 must fetch, build, and maintain slot state<sup>3</sup><br>🟡 less chance of startup delays<sup>4</sup><br>🔴 may slow Codex client during startup<sup>5</sup><br>🔴 must run one validator per bucket to cover entire space<sup>6</sup><br>🔴 additional CLI param needed<sup>14</sup><br>🔴 additional Codex client complexity<sup>16</sup> | 🟡 less chance of validation delays<sup>1,2</sup><br>🔴 increased gas requirement for `fillSlot`<sup>7</sup><br>🟡 less chance of hitting RPC limits<sup>8</sup><br>🔴 must run one validator per bucket to cover entire space<sup>6</sup><br>🟢 no startup delays<sup>9</sup><br>🔴 change to validator buckets requires contract re-deploy<br>🔴 additional CLI param needed<sup>14</sup> |
| No slot buckets | 🔴 cannot guarantee `SlotId` space is covered<sup>10</sup><br>🔴 increased chance of validation delays<sup>1,11</sup><br>🔴 increased chance of startup delays<sup>4</sup><br>🔴 must fetch, build, and maintain slot state<sup>3</sup><br>🟢 simplifies CLI params<sup>12</sup><br>🟡 must run one validator covering all slots<sup>13</sup><br>🔴 additional Codex client complexity<sup>16</sup>                                        | 🔴 cannot guarantee `SlotId` space is covered<sup>10</sup><br>🔴 increased chance of validation delays<sup>1,11</sup><br>🔴 increased gas requirement for `fillSlot`<sup>7</sup><br>🔴 increased chance of hitting RPC limits<sup>15</sup><br>🟢 no startup delays<sup>9</sup>                                                                                                                                              |


<sup>1</sup> The number of slots to validate may exceed the validators' capabilities to validate all required slots in a single period. If this happens, then SPs may be able to get paid out at the end of contracts before validators have had a chance to validate that they didn't miss proofs. A potential solution for this, is to disallow `freeSlot` to be called for some periods after the contract has finished.

<sup>2</sup> Assumes configured number of validators is large enough to keep the bucket size to within the validators' capacity to validate slots within a single period. The chance of exceeding validator capabilities is less than if there were no slot buckets, due to less slots required for validation.

<sup>3</sup> Slot state maintenance has a few disadvantages:
1.    fetching and building the slot state locally increases the startup time for the validator, with a fresh start being potentially very long.
2.    increased memory footprint
3.    must consider the complexities of (single-threaded) concurrency issues with simultaneous reading and writing of the slot state, similar to those experienced with the slot queue

<sup>4</sup> Before a validator begins validating, it must check the current state of the slot. The more slots to validate, the longer it will take to start validating slots at startup. With buckets in place, the number of slots will be less than without buckets.

<sup>5</sup> Building the slot state may rob needed CPU/IO from other parts of the Codex startup routines

<sup>6</sup> To ensure that the entire SlotId space is covered by validators, a minimum of one validator per bucket must be running at all times.

<sup>7</sup> `fillSlot` stores the `SlotId` in the corresponding bucket, adding to the gas costs. However, the gas costs may decrease for `freeSlot` and `payoutSlot` as storage is being freed.

<sup>8</sup> Compared to fetching all `SlotIds` when not limited in number by buckets

<sup>9</sup> Compared to fetching and building the slot state locally

<sup>10</sup> The entirety of the `SlotId` cannot be guaranteed to be validated unless there is one validator that is running that can handle validation of all the slots in the network

<sup>11</sup> With no slot buckets to limit the number of slots to validate, there is an increased chance the validator will not be able to validate all slots in the network in a given period.

<sup>12</sup> No additional validator CLI params are needed, only `--validator-max-slots` is needed which is already implemented.

<sup>13</sup> No need to consider if one validator for each bucket are running, but must ensure that one validator that can handle all slots in the network is running.

<sup>14</sup> When slot buckets are used, an additional CLI param is required to assign the validator to a particular bucket. If not specified, by default a random bucket could be assigned, or slots in all buckets could be assigned. In both cases, `--validator-max-slots` would limit the slots validated.

<sup>15</sup> Compared to fetching `SlotIds` when limited in number by buckets

<sup>16</sup> The complexity required for the local slot state structure is more complex and will take more time to implement and maintain than storing it on chain.

emizzle:
After a discussion, it was highlighted that in order to prevent SPs from being locked into lengthy contracts, a maximum duration of initially 30 days will be implemented. Due to this maximum, querying and processing 30 days of historical `SlotFilled` events is sufficient to cover active slots in the network. On startup, Validators will issue one RPC request to retrieve all `SlotFilled` events over the past 30 days and build the list of `SlotIds` in memory. Only `SlotIds` that are in the assigned bucket will be added to the in-memory list. The validator will then process the `SlotIds` as they are processed now, in a loop, removing slots that are no longer filled.

@markspanbroek @AuHau, I would like to point out, that even though the number of slots is limited to slots filled in the past 30 days, and limited to the assigned bucket, the number could be substantial and take quite a lot of time to check the state of each slot in the `removeSlotsThatHaveEnded` routine. Empirically, in our current testnet setup, queries to the RPC server for fetching the slot state take on the order of 300ms (due to my physical distance from the RPC server). If there are 1000 slots in the list, it will take 300 seconds (5 minutes) for the validator to complete one iteration of `removeSlotsThatHaveEnded`. 

Did we figure out what we decided to do if there are too many slots for the validator to be able to validate in a given period?

emizzle:
> Did we figure out what we decided to do if there are too many slots for the validator to be able to validate in a given period?

Add an error log if this occurs.

Another point, is that the validator does not have an entire period to validate all slots. It has from the end of the last period to `timeout` seconds to complete validation.

AuHau:
@emizzle I believe this PR can be closed as we are going with the local approach right?

emizzle:
Closing in favour of https://github.com/codex-storage/nim-codex/pull/890

Issue #112, opened 2024-05-31
Add Codex Testnet network


veaceslavdoina:
As we discussed, we  should introduce `testnet` for Codex Testnet and that require some modificaitons
 - Add new network to the [`hardhat.config.js`](https://github.com/codex-storage/codex-contracts-eth/blob/master/hardhat.config.js)
 - Add new network to the [`verifier/networks`](https://github.com/codex-storage/codex-contracts-eth/tree/master/verifier/networks)

We discussed the name `testnet`, but shouldn't `codex_testnet` be more appropriate similar to `taiko_test`?

Part of https://github.com/codex-storage/infra-codex/issues/186

emizzle:
> shouldn't codex_testnet be more appropriate similar to taiko_test?

Since `taiko_test` deployed to the Taiko L2 network, it makes sense for it to be named as such. In our case, the network is our own test Ethereum network, so we could name it something, as is typical for these types of testnets, eg `rinkeby` is one of Ethereum's testnets. Have [asked the team](https://discord.com/channels/895609329053474826/1230457525854539819/1246981358849359983) what they think regarding suitable names.

PR #135, opened 2024-07-31
Add Codex Devnet and Testnet networks

veaceslavdoina:
This RP implements #112 and a part of https://github.com/codex-storage/infra-codex/issues/192

As we discussed, we plan to run soon a Devnet and Testnet. In order to be able to deploy and store artifacts for different networks, we would need separate configurations for them in [`hardhat.config.js`](https://github.com/codex-storage/codex-contracts-eth/blob/master/hardhat.config.js).

We propose to use the following approach
| Network       | Contracts network |
| ------------- | ----------------- |
| Codex Devnet  | `codex_devnet`    |
| Codex Testnet | `codex_testnet`   |
| Codex Mainnet | `codex_mainnet`   |

And template it as follow - `<vendor>_<network>[_custom]`
> **Note:** We use `_` as a separator like that is done for `taiko_test`

In that way we can extend networks like following
| Network       | Contracts network  | Description      |
| ------------- | ------------------ | ---------------- |
| Codex Devnet  | `codex_devnet`     |                  |
| Codex Devnet  | `codex_devnet_pos` | Codex Devnet PoS |

If you have a different proposal let's discuss and implement it and move forward with Devnet/Testnet deployments.

We can rename `codexdisttestnetwork` later to the `codex_disttests` or `codex_dist-tests` to follow an unified approach.


gmega:
I believe, sadly, that you'll also have to replicate the folders in the `verifier` subfolder. I believe we'll need a set of circuit assets per deployment, taking the size of this repo to a nice 200 MB+.

gmega:
Btw guys, for context: I had a chat in private with @veaceslavdoina the other day and we're trying to get something simple out:

* devnet tracks master;
* testnet tracks some controlled release process (manual deploys of tags).

mainnet is future.

veaceslavdoina:
Added circuit assets based on the [latest build](https://github.com/codex-storage/codex-storage-proofs-circuits/actions/runs/9906695917/job/27368791818)
```shell
# Directory
cd verifier/networks

# Download
curl https://circuit.codex.storage/proving-key/852bdc217fedf216ac9e85f9f9aa86964e4d66e8edfddc9a4cdf7c764990c857 -O

# Unzip
unzip -d codex_devnet 852bdc217fedf216ac9e85f9f9aa86964e4d66e8edfddc9a4cdf7c764990c857
unzip -d codex_testnet 852bdc217fedf216ac9e85f9f9aa86964e4d66e8edfddc9a4cdf7c764990c857

# Cleanup
rm -f 852bdc217fedf216ac9e85f9f9aa86964e4d66e8edfddc9a4cdf7c764990c857
```

Current size
```shell
du -hs verifier/networks

369M	verifier/networks
```

Warning from GitHub
Screenshot 2024-07-31 at 16 07 43

veaceslavdoina:
In the [codexdisttestnetwork](https://github.com/codex-storage/codex-contracts-eth/tree/master/verifier/networks/codexdisttestnetwork) we have also a [`zkey_hash.json`](https://github.com/codex-storage/codex-contracts-eth/blob/master/verifier/networks/codexdisttestnetwork/zkey_hash.json) file, but is not in the downloaded assets - how we can get it?

gmega:
I'm unaware of how those artifacts you're downloading are produced, @veaceslavdoina, but the ones uploaded from the codex proof circuit build (e.g. [like this build](https://github.com/codex-storage/codex-storage-proofs-circuits/actions/runs/9906695917/job/27368791818)) contain the `zkey_hash.json` and the circom main module inside of the zip file.

emizzle:
Regarding naming, I think we need to cover the following networks:
```
codex_devnet
codex_testnet       # we should avoid using this imo, as it is not versioned
codex_testnet_v1 # numbered version testnet, if we don't use named versions
codex_crossbow   # named version testnet, if we don't use numbered versions
optimism_goerli
optimism_sepolia
optimism_mainnet
zksync_sepolia
zksync_mainnet
```
Abstracting that, I think the following pattern is probably sufficient:
```
<network>_<version>
```
I'm not convinced we need the `_custom` suffix. For the numbered version, eg `codex_testnet_v1`, `<version>` could be `testnet_v1`. And I don't think we'll be running our own PoS network, although I understand this was just an arbitrary example.

Another note, my opinion is we should probably avoid using `codex_testnet`, as it is not versioned, so we should probably decide to version it with a number or a name. I'm fine with either, though I lean towards a number since there is something concrete that can be matched up with a branch or a tag. I guess our release branches/tags could also be named, but then we may have a more difficult time knowing what the order is.

It's probably obvious, but I think it's important to point out that besides the need to generate a new set of ceremony assets for each network, these networks only really specify an RPC endpoint of where to deploy the contracts to. As such, I don't think we're going to need a `codex_mainnet` network, since for our mainnet deployment, the network to deploy the contracts to will be an L2 network like `optimism_mainnet` or `zksync_mainnet`.

veaceslavdoina:
Just run 2 additional assets generations
- [codex_devnet](https://github.com/codex-storage/codex-storage-proofs-circuits/actions/runs/10191981826)
- [codex_testnet](https://github.com/codex-storage/codex-storage-proofs-circuits/actions/runs/10192013007)

And updated assets in the repository accordingly, but only 3 files are different
**different**
```
proof_main.zkey
proof_main_verification_key.json
zkey_hash.json
```
**same**
```
proof_main.circom
proof_main.r1cs
proof_main.wasm
```

Will rename networks per Eric's proposal if we end up with that approach.

markspanbroek (approved):
Looks good to me. We can always change naming later? We can even rename codex_testnet to codex_testnet_v1 after deployment, I think.

emizzle (approved):
Let's go with what we have in this PR. It will support named networks, so should work for all the different configs I suggested earlier. 

Worst case, we can update it later. Definitely better to get something out than overdiscuss

PR #138, opened 2024-08-05
chore: add `downtimeProduct` configuration parameter

### LINKED ISSUE #108: Period multiplicator as part of the Proof's configuration
https://github.com/codex-storage/codex-contracts-eth/blob/master/contracts/Proofs.sol#L48

The Period multiplicator (67) in `getPointer()` function should be configurable as it is coupled with the downtime parameter.

We should also better understand what configuration values we will use in real deployment, because then maybe we would not even need the multiplication? I would assume that the Period should be quite long, but Mark pointed out that it should not be longer than the 256 blocks window.

emizzle:
Closes #108.


AuHau (approved):

markspanbroek (approved):
:+1:

PR #140, opened 2024-08-05
fix: ensure requestStorage() reverts if maxSlotloss > slots

### LINKED ISSUE #139: Check that `maxSlotLoss` is not bigger then `slots`
There should be this check in `fillSlot`: https://github.com/codex-storage/codex-contracts-eth/blob/master/contracts/Marketplace.sol#L113

0x-r4bbit:
`request.ask.maxSlotLoss` has to be less or equal to `slots`. This check was missing, so this commit adds it and provdes a test to confirm this as well.

Fixes #139


AuHau (approved):
LGTM, thanks 🙏

Just fix your formatting ;-)

0x-r4bbit:
Done

PR #141, opened 2024-08-05
refactor(Marketplace): Use custom errors instead of string messages

0x-r4bbit:
This reduces the gas cost as custom errors always use 4 bytes.

This supersedes #73 as we need PRs to be sent from upstream repository in order to have the verification task work on CI.


deployments/taiko_test/solcInputs/9671d4e1820229dcd69b349a99feb65b.json
```diff
@@ -62,4 +62,4 @@
       "useLiteralContent": true
     }
   }
-}
\ No newline at end of file
+}
```
emizzle, line 65: Was this manually added? If so, maybe we should leave it out as I believe this is a generated file, so we shouldn't modify it.

emizzle (approved):
LGTM

contracts/Marketplace.sol
```diff
@@ -87,18 +107,20 @@ contract Marketplace is Proofs, StateRetrieval, Endian {
...
-      "Expiry not in range"
-    );
-    require(
-      request.ask.maxSlotLoss <= request.ask.slots,
-      "maxSlotLoss exceeds slots"
-    );
+    if (_requests[id].client != address(0)) {
+      revert Marketplace_RequestAlreadyExists();
+    }
+    if (request.expiry == 0 || request.expiry > request.ask.duration) {
```
markspanbroek, line 118: There is a subtle change here, where request.expiry == request.ask.duration was previously not allowed, but it is allowed now. Not a problem, but something to keep in mind when integrating this into nim-codex. Maybe there's an integration test that depends on the previous behaviour.

markspanbroek (approved):
:+1:

contracts/Marketplace.sol
emizzle: It's probably better to change it back, imo.

AuHau:
Btw. please don't merge it until we have `nim-codex` PR that will be implementing this. I assume you don't wanna dive in Nim, so I will have a look on it in the following days.

contracts/Marketplace.sol
0x-r4bbit: Very good catch, I'm gonna change this.

deployments/taiko_test/solcInputs/9671d4e1820229dcd69b349a99feb65b.json
0x-r4bbit: I believe this is a formatting thing from prettier. I haven't touched the file myself.
Will remove this change from the commit entirely

0x-r4bbit:
Addressed all comments made here and updated the PR.

0x-r4bbit:
At last!

PR #142, opened 2024-08-05
chore: formally verify request state changes

### LINKED ISSUE #128: Implement application property: only specific state transitions are allowed
This issue is about implementing the application property that only certain state transitions of `SlotState` and `RequestState` are allowed.

0x-r4bbit:
This commit adds CVL rule that formally verifies the state changes of
any given request in relation to the functions of the contract that can
cause them.

Closes #128

### DIFF
--- a/certora/harness/MarketplaceHarness.sol
+++ b/certora/harness/MarketplaceHarness.sol
@@ -1,0 +1,18 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.23;
+
+import {IERC20} from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import {IGroth16Verifier} from "../../contracts/Groth16.sol";
+import {MarketplaceConfig} from "../../contracts/Configuration.sol";
+import {Marketplace} from "../../contracts/Marketplace.sol";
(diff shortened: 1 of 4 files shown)


certora/harness/MarketplaceHarness.sol
```diff
@@ -0,0 +1,18 @@
...
+import {MarketplaceConfig} from "../../contracts/Configuration.sol";
+import {Marketplace} from "../../contracts/Marketplace.sol";
+import {RequestId} from "../../contracts/Requests.sol";
+
+contract MarketplaceHarness is Marketplace {
+    constructor(MarketplaceConfig memory config, IERC20 token, IGroth16Verifier verifier) Marketplace(config, token, verifier) {}
+
+    function requestContext(RequestId requestId) public returns (Marketplace.RequestContext memory) {
+        return _requestContexts[requestId];
+    }
```
0x-r4bbit, line 16: Atm this harness is used instead of the actual contract, so we get access to `_requestContexts`.

Looking into solving this with `ghost` variables instead so we don't have to add this harness and change the property's visibility.

AuHau (approved):

Issue #71, opened 2023-10-08
Reward address


AuHau:
As Ethereum 2 validators have several addresses, maybe the Codex node could have a "reward address"?

AuHau:
https://ethereum.org/en/staking/withdrawals/

emizzle:
Provide payout/profit address on the `freeSlot` routine.

emizzle:
Profits to reward address, but collateral gets returned to the address from which it came?

AuHau:
> Profits to reward address, but collateral gets returned to the address from which it came?

Yes

PR #144, opened 2024-08-07
feat: adds an optional `payoutAddress` to allow payouts to be paid to separate address

emizzle:
Closes: #71.

Adds an optional `payoutAddress` to the `fillSlot` function, tracking this address in the `Slot`. When SPs are paid out after request completion/cancellation, reward payouts are paid to the `payoutAddress` (if supplied), while collateral is returned to the SP's address. If the `payoutAddress` is not supplied in `fillSlot`, payouts are made to the SP's address (as it was before this PR).

### DIFF
--- a/contracts/Marketplace.sol
+++ b/contracts/Marketplace.sol
@@ -116,25 +116,6 @@ contract Marketplace is Proofs, StateRetrieval, Endian {
   }
 
   /**
-   * @notice Fills a slot with the payout address set to the address of the host
-     filling the slot.
-   * @param requestId RequestId identifying the request containing the slot to
-     fill.
-   * @param slotIndex Index of the slot in the request.
-   * @param proof Groth16 proof procing possession of the slot data.
-   * @dev This overload is an alternative way of providing an optional parameter
-     for payoutAddress, used for instances where a host does not specify a
-     separate payout address.
-   */
-  function fillSlot(
-    RequestId requestId,
-    uint256 slotIndex,
-    Groth16Proof calldata proof
-  ) public requestIsKnown(requestId) {
-    fillSlot(requestId, slotIndex, proof, msg.sender);
-  }
-
-  /**
    * @notice Fills a slot. Reverts if an invalid proof of the slot data is
      provided.
    * @param requestId RequestId identifying the request containing the slot to
--- a/contracts/Marketplace.sol
+++ b/contracts/Marketplace.sol
@@ -46,7 +46,8 @@ contract Marketplace is Proofs, StateRetrieval, Endian {
     /// @dev When Slot is filled, the collateral is collected in amount of request.ask.collateral
     /// @dev When Host is slashed for missing a proof the slashed amount is reflected in this variable
     uint256 currentCollateral;
-    address host;
+    address host; // address used for collateral interactions and identifying hosts
+    address payoutAddress; // address used for receiving payouts
   }
 
   struct ActiveSlot {
@@ -117,7 +118,8 @@ contract Marketplace is Proofs, StateRetrieval, Endian {
   function fillSlot(
     RequestId requestId,
     uint256 slotIndex,
-    Groth16Proof calldata proof
+    Groth16Proof calldata proof,
+    address payoutAddress
   ) public requestIsKnown(requestId) {
     Request storage request = _requests[requestId];
     require(slotIndex < request.ask.slots, "Invalid slot");
@@ -133,6 +135,7 @@ contract Marketplace is Proofs, StateRetrieval, Endian {
     submitProof(slotId, proof);
 
     slot.host = msg.sender;
+    slot.payoutAddress = payoutAddress;
     slot.state = SlotState.Filled;
     slot.filledAt = block.timestamp;
     RequestContext storage context = _requestContexts[requestId];
@@ -171,6 +174,7 @@ contract Marketplace is Proofs, StateRetrieval, Endian {
     } else if (state == SlotState.Failed) {
       _removeFromMySlots(msg.sender, slotId);
     } else if (state == SlotState.Filled) {
+      // free slot without returning collateral, effectively a 100% slash
       _forciblyFreeSlot(slotId);
     }
   }
@@ -231,6 +235,13 @@ contract Marketplace is Proofs, StateRetrieval, Endian {
     }
   }
 
+  /**
+   * @notice Abandons the slot without returning collateral, effectively slashing the
+     entire collateral.
+   * @param slotId SlotId of the slot to free.
+   * @dev _slots[slotId] is deleted, resetting _slots[slotId].currentCollateral
+     to 0.
+  */
   function _forciblyFreeSlot(SlotId slotId) internal {
     Slot storage slot = _slots[slotId];
     RequestId requestId = slot.requestId;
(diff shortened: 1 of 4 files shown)


AuHau (commented):
Good job! Generally LGTM. Thanks for adding all the comments!

My only question is: should we allow not specifying the payout address? In my mind, when I was thinking about this, I automatically assumed that it would be a mandatory thing.

emizzle:
> My only question is: should we allow not specifying the payout address? In my mind, when I was thinking about this, I automatically assumed that it would be a mandatory thing.

My preference is to allow the fallback address to be the host's address if nothing is specified. It is one less *required* step for a SP to get up and running. But I do see that requiring this address is forcing good opsec on the SP, allowing them to set their host address if they really wanted to 🤔

AuHau:
> But I do see that requiring this address is forcing good opsec on the SP, allowing them to set their host address if they really wanted to 🤔

Yeah exactly. Lets touch on this today on the call.

contracts/Marketplace.sol
```diff
@@ -270,30 +310,45 @@ contract Marketplace is Proofs, StateRetrieval, Endian {
...
-    uint256 amount = _requests[requestId].pricePerSlot() +
-      slot.currentCollateral;
-    _marketplaceTotals.sent += amount;
+    uint256 payoutAmount = _requests[requestId].pricePerSlot();
+    uint256 collateralAmount = slot.currentCollateral;
+    _marketplaceTotals.sent += (payoutAmount + collateralAmount);
     slot.state = SlotState.Paid;
-    assert(_token.transfer(slot.host, amount));
+    assert(_token.transfer(slot.payoutAddress, payoutAmount));
+    assert(_token.transfer(slot.host, collateralAmount));
```
0x-r4bbit, line 318: Not something that has to be done in this PR, but we should consider using `require()` instead of `assert()` as `assert()` [eats all the remaining gas](https://www.evm.codes/#fe?fork=cancun).

0x-r4bbit (commented):
I'd also vote for making the payment address mandatory.
Usually in systems where some sort of payout can happen, the protocol should ask where the payment should go, because the `msg.sender` address could be blacklisted in tokens, preventing it from getting paid.

contracts/Marketplace.sol
emizzle: Good catch! Created an issue to track this: https://github.com/codex-storage/codex-contracts-eth/issues/149

emizzle:
`payoutAddress` is now required for `fillSlot`. However, non-optionality *may* not solve the issue that @0x-r4bbit pointed out:

> Usually in systems where some sort of payout can happen, the protocol should ask where the payment should go, because the msg.sender address could be blacklisted in tokens, preventing it from getting paid.

This is because returned collateral is still sent to `slot.host` and `slot.host` is assigned using `msg.sender` in `fillSlot`:
https://github.com/codex-storage/codex-contracts-eth/blob/6db5479c855ff9048c6557aa7966a40d308492b4/contracts/Marketplace.sol#L299
https://github.com/codex-storage/codex-contracts-eth/blob/6db5479c855ff9048c6557aa7966a40d308492b4/contracts/Marketplace.sol#L147

Thoughts @0x-r4bbit?

markspanbroek:
Perhaps you can add an optional payoutAddress to freeSlot() instead? Then you also don't need to store it.

emizzle:
> Perhaps you can add an optional payoutAddress to freeSlot() instead? Then you also don't need to store it.

Good point. Regarding the optional parameter, what would happen if a payout address wasn't provided?

0x-r4bbit:
> For me, it is not clear where msg.sender blacklisting would happen?

@emizzle Token contracts can black list addresses, as well as exchanges. You want to be able to send your rewards/payouts to an address that is not affected by anything.

markspanbroek:
> Regarding the optional parameter, what would happen if a payout address wasn't provided?

I'd use `slot.host` if the payout address wasn't specified. The reason for having the parameter optional, is that in some cases there is no payout (when the storage request failed for instance).

AuHau:
Good idea about the `freeSlot()` but I would still leave the payout address parameter mandatory. It is not like you don't know it when the node calls it even for the failed addresses and as Eric pointed out, if you really want to use `msg.sender` then you can specify it in the client CLI input.

emizzle:
> Token contracts can black list addresses, as well as exchanges. You want to be able to send your rewards/payouts to an address that is not affected by anything.

That is a good point. We are still using `msg.sender` to transfer tokens during a withdraw (when a client's request is cancelled, they can withdraw the remaining funds). I wonder if it's worth adding a separate parameter in `withdrawFunds` to mitigate any issues with blacklisting? Thoughts?

emizzle:
Moved `payoutAddress` from `fillSlot` to `freeSlot`, and removed it from the `Slot` struct (no longer needed for storage).

0x-r4bbit:
> . I wonder if it's worth adding a separate parameter in withdrawFunds to mitigate any issues with blacklisting?

@emizzle I'd say it's worth considering for sure. Curious about @0xb337r007 's thoughts on this.

markspanbroek (approved):
Looks good to me!

I think it's a good idea to make `withdrawFunds()` work the same way, it's more consistent.

AuHau (approved):
LGTM, thanks! Also agree with adding the same functionality to `withdrawFunds()`

emizzle:
There is still an issue: collateral is returned to the SP, which is a different address than the payout address. Should we  change the behaviour so that collateral is also returned to the payout address (to avoid blacklisting)?

AuHau:
Good catch Eric. 

Hmm, what about adding `collateralAddress` (or similar)? That would allow specifing where the collateral should be sent? So we could support the idea of having "split piles" for the operational funds (eq. collateral) and gains (eq. rewards). 

And if somebody would get blacklisted and would like to retrieve collateral then he would have to manually create the transaction with different addresses? I would not try to support this edge case in the client for now, but it would be good to have it in the contract IMHO.

markspanbroek:
I like your idea Adam, allowing two addresses to be specified in freeSlot. I would then consider to add an overloaded version that takes only one argument, and sends the collateral back to the host address. So that if you want to call freeSlot() and don't care about this edge case, you also don't have to think about it.

emizzle:
After my third round of completely redoing this 😂 , I'd like to revisit the motivation for this task so that we can coalesce on a solution.

### Avoid key leaks
Motivation: by specifying a reward payout address, we can protect profits from the vulnerability of private key memory leaks (of the SP's main wallet).

We can avoid key leaks by allowing an SP to specify a payout address using `--reward-recipient` (or `--payout-address` or whatever we call it) in the CLI and passing this to `freeSlot`. The more I think about this, the more I like the idea of keeping the reward recipient as an optional address. It allows an SP to get started with less configuration and allows the flexibility of increased OpSec or increased convenience. A singular address with collateral and profit distribution means less interactions are needed if an SP wants to use profits to engage with more storage requests.

### Circumvent token black listings
Motivation:
> Token contracts can black list addresses, as well as exchanges. You want to be able to send your rewards/payouts to an address that is not affected by anything.

AFAIK, the Codex token will likely not have a blacklist and won't be using an existing token (eg a stablecoin), so `SP <-- transfer --> token` should be valid even if SP's address is black listed *somewhere else*. This reduces the issue surface to SP interactions with a CEX/DEX being blocked. This can be worked around by the SP sending their tokens to another address before interacting with a CEX/DEX. Altogether, I think this point is mostly moot, and allowing client withdraws / SP collateral refunds to separate addresses is merely a convenience. It makes sense to allow `withdrawFunds` and `freeSlot` to have _optional_ parameters that allow paying out to these addresses in the case of a black listing, however, I don't think they should be implemented in the client, to avoid CLI parameter overload and confusion.

### Changes summary
To sum up the changes that this would introduce, we'd have:

`withdrawFunds`
  - `address withdrawRecipient` (optional): address to withdraw the remaining client funds to, defaults to SP's address
    - No client CLI option

`freeSlot`
  - `address rewardRecipient` (optional): address to send SP reward payouts to, defaults to SP's address
    - Uses `--reward-recipient` CLI option in the client
  - `address collateralRecipient` (optional): address to send SP collateral refunds to, defaults to SP's address
    - No client CLI option

AuHau:
> After my third round of completely redoing this 😂 , I'd like to revisit the motivation for this task so that we can coalesce on a solution.

Yeah, this suck - sorry about that :-( 

> It allows an SP to get started with less configuration.

Hmm, that is good point I have not considered. Maybe then we could "handle" this on a documentation level (when we have one 😅) with some big warning box that from OpSec it is better to have those account separated, but leave it optional. Agree 👍

Agree with the changes summary - nicely written!

markspanbroek:
I also agree with your changes summary @emizzle, thanks! :+1:

PR #147, opened 2024-08-09
chore(certora): add invariant that totalSent is <= totalReceived

### LINKED ISSUE #132: Implement application property: marketplaceTotals.sent <= maketplaceTotals.received
This issue is about implementing the application property that the marketplace totals `sent` value is always <= its totals `received` value.

0x-r4bbit:
This commit adds an invariant that verifies `marketplaceTotals.sent <= marketplceTotals.received`.

The invariant relies on the `Token.totalSupply()` which has previously been verified to be the sum of all balances.

The invariant uses `<=` as there could be donations sent to the marketplace.

Closes #132

**This PR depends on #146**

### DIFF
--- a/certora/specs/Marketplace.spec
+++ b/certora/specs/Marketplace.spec
@@ -253,6 +253,22 @@ invariant cancelledSlotAlwaysHasCancelledRequest(env e, Marketplace.SlotId slotI
     currentContract.slotState(e, slotId) == Marketplace.SlotState.Cancelled =>
         currentContract.requestState(e, slotIdToRequestId[slotId]) == Marketplace.RequestState.Cancelled;
 
+// STATUS - in progress (fails on constructor state due to a strange behaviour in the tool)
+// _marketplaceTotals.received - _marketplaceTotals.sent == tokenBalanceOfContract
+// https://prover.certora.com/output/3106/4e6cac8055ac45bb92840e14d9b095eb/?anonymousKey=655a42ca6306a023db78914d5a188d8ec2882771
+invariant totalSentIsLessThanOrEqualTotalReceived()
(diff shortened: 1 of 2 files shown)


certora/specs/Marketplace.spec
```diff
@@ -72,6 +72,22 @@ invariant totalSupplyIsSumOfBalances()
 invariant requestStartedWhenSlotsFilled(env e, Marketplace.RequestId requestId, Marketplace.SlotId slotId)
     to_mathint(currentContract.requestContext(e, requestId).slotsFilled) == to_mathint(currentContract.getRequest(e, requestId).ask.slots) => currentContract.requestState(e, requestId) == Marketplace.RequestState.Started;
 
+// STATUS - in progress (fails on constructor state due to a strange behaviour in the tool)
+// _marketplaceTotals.received - _marketplaceTotals.sent == tokenBalanceOfContract
+// https://prover.certora.com/output/3106/4e6cac8055ac45bb92840e14d9b095eb/?anonymousKey=655a42ca6306a023db78914d5a188d8ec2882771
```
0x-r4bbit, line 77: As pointed out here, there's a bug in the prover that causes this invariant to file after the constructor has been called.

I'd leave this pending until the bug in certora has been fixed.

AuHau (approved):

0x-r4bbit:
While this is green on CI, it should actually fail because there's a bug in the prover that will find a counter example on one of the rules.

We need to change the GH action in this repo so it has a proper exit code or something.

AuHau:
As per https://github.com/codex-storage/codex-contracts-eth/issues/132#issuecomment-2260669669 we won't merge this.

PR #152, opened 2024-08-13
chore(certora): verify SlotState and RequestState changes count

### LINKED ISSUE #129: Implement application property: state can't transition more than once per function call
This issue is about implementing the application property that a state transition of `SlotState` and `RequestState` cannot happen more than once in any given function call of the contract.

0x-r4bbit:
This adds a rule to formally verify that the `SlotState` and `RequestState` of any given `Slot` or `Request` does not change more than once per function call.

Closes #129


0x-r4bbit:
This rule was also reviewed by certora.

AuHau (approved):
LGTM! Thanks!

PR #155, opened 2024-08-14
chore(certora): slot's missed periods count should be equal to the count of slot's missing periods set to true

### LINKED ISSUE #133: Implement application property: number of `_missed` is <= number of periods with missd proofs
This issue is about implementing the application property that the number of `_missed` proofs for certain slot IDs is less or equal to the number of periods with missed proofs.

gravityblast:
This invariant makes sure that `_missed[slotId]` is equal to the count of `_missing[slotId]` periods set to `true`.

Closes #133


0x-r4bbit (approved):
LGTM!

0x-r4bbit:
Task on CI is also green

PR #160, opened 2024-08-19
feat: hosts payed by actual time hosting a slot

### LINKED ISSUE #98: Send client remaining funds when a contract fails
When a contract completes (succeeds or fails), send client remaining funds.

### LINKED ISSUE #100: Proportional payout for repaired slots
Currently, when a slot is freed and another host repairs it, they receive the repair reward. Upon successfully completing the slot, they will receive the reward for the entire storage contract duration, not just a portion based on how long they filled the slot (e.g., from the time of slot repair).

Is this desirable? Could it pose a problem?

### LINKED ISSUE #102: Payout based on actual time filling the slot
Currently, if a node fills a slot and the request successfully finishes, it gets paid for the **entire request duration** regardless of when the node actually filled the slot. 

In the context of Slot Reservations, we proposed that changing this behavior to payment based on the actual time of slot filling would create a small incentive to fill the slot as soon as possible.

### LINKED ISSUE #156: Return funds to client for failed request
Client should be refunded remaining funds after a contract fails.

AuHau:
This PR bundles few changes around payout and withdraws:

 - Hosts are paid by their actual time hosting the slot (also relevant for repair)
 - Client can withdraw remaining funds for finished requests
 - Client can withdraw all original funds for failed requests
 - When a slot is "forcible freed" and the slot would not be filled again, but the request would successfully finish - client can withdraw the funds for this slot
 
I am not so happy about the testing because it is pretty hard to correctly calculate the amounts that should be returned due to Hardhat's "auto" mining of blocks every time there is a transaction, which advances the clock and hence shifts the reward calculations. I made it work, but not great. My brain is fried though, so I will leave it for now and check back on it tomorrow. If you have any suggestions on how to track the funds better, I would love to hear them!

Closes #98
Closes #100 
Closes #102 
Closes #156

### DIFF
--- a/contracts/Marketplace.sol
+++ b/contracts/Marketplace.sol
@@ -386,19 +386,16 @@ contract Marketplace is SlotReservations, Proofs, StateRetrieval, Endian {
     Request storage request = _requests[requestId];
     require(request.client == msg.sender, "Invalid client address");
     RequestContext storage context = _requestContexts[requestId];
+    RequestState state = requestState(requestId);
     require(
-      context.state == RequestState.New ||
-        context.state == RequestState.Failed ||
-        requestState(requestId) == RequestState.Finished,
+      state == RequestState.Cancelled ||
+        state == RequestState.Failed ||
+        state == RequestState.Finished,
       "Invalid state"
     );
     require(context.fundsToReturnToClient != 0, "Nothing to withdraw");
 
-    if (context.state == RequestState.New) {
-      require(
-        block.timestamp > requestExpiry(requestId),
-        "Request not yet timed out"
-      );
+    if (state == RequestState.Cancelled) {
       context.state = RequestState.Cancelled;
       emit RequestCancelled(requestId);
 
@@ -409,7 +406,7 @@ contract Marketplace is SlotReservations, Proofs, StateRetrieval, Endian {
       context.fundsToReturnToClient +=
         context.slotsFilled *
         _payoutAmount(requestId, requestExpiry(requestId));
-    } else if (context.state == RequestState.Failed) {
+    } else if (state == RequestState.Failed) {
       // For Failed requests the client is refunded whole amount.
       context.fundsToReturnToClient = request.maxPrice();
     } else {
@@ -458,7 +455,6 @@ contract Marketplace is SlotReservations, Proofs, StateRetrieval, Endian {
     if (state == RequestState.New || state == RequestState.Started) {
       return end;
     } else {
-      /// For F
       return Math.min(end, block.timestamp - 1);
     }
   }
--- a/contracts/Marketplace.sol
+++ b/contracts/Marketplace.sol
@@ -30,8 +30,7 @@ contract Marketplace is SlotReservations, Proofs, StateRetrieval, Endian {
     RequestState state;
     uint256 slotsFilled;
     /// @notice Tracks how much funds should be returned to the client as not all funds might be used for hosting the request
-    /// @dev The sum starts on the full reward amount for the request and is deducted every time a host fills a Slot by precalculated amount that the host should receive
-    ///  if the he successfully finished the request (eq. for the time between when he filled the slot and request's end).
+    /// @dev The sum starts with the full reward amount for the request and is reduced every time a host fills a slot. The reduction is calculated from the duration of time between the slot being filled and the request's end. This is the amount that will be paid out to the host when the request successfully finishes.
     uint256 fundsToReturnToClient;
     uint256 startedAt;
     uint256 endsAt;
@@ -42,7 +41,7 @@ contract Marketplace is SlotReservations, Proofs, StateRetrieval, Endian {
     SlotState state;
     RequestId requestId;
     /// @notice Timestamp that signals when slot was filled
-    /// @dev Used for calculating payouts as Hosts are payed based on time they actually host the content
+    /// @dev Used for calculating payouts as hosts are paid based on time they actually host the content
     uint256 filledAt;
     uint256 slotIndex;
     /// @notice Tracks the current amount of host's collateral that is to be payed out at the end of Slot's lifespan.
@@ -365,7 +364,7 @@ contract Marketplace is SlotReservations, Proofs, StateRetrieval, Endian {
   /**
    * @notice Withdraws remaining storage request funds back to the client that
      deposited them.
-   * @dev Request must be expired (hence be in RequestStat e.New), failed or finished, and the
+   * @dev Request must be cancelled, failed or finished, and the
      transaction must originate from the depositor address.
    * @param requestId the id of the request
    */
@@ -403,9 +402,10 @@ contract Marketplace is SlotReservations, Proofs, StateRetrieval, Endian {
       context.state = RequestState.Cancelled;
       emit RequestCancelled(requestId);
 
-      // The fundsToReturnToClient tracks funds to be returned for requests that succesfully finishes
-      // When request get cancelled, we need to refund the funds which should have payed for the duration
-      // between <expiresAt;endsAt> for every slot that was filled.
+      // `fundsToReturnToClient` currently tracks funds to be returned for requests that successfully finish.
+      // When requests are cancelled, funds earmarked for payment for the duration
+      // between request expiry and request end (for every slot that was filled), should be returned to the client.
+      // Update `fundsToReturnToClient` to reflect this.
       context.fundsToReturnToClient +=
         context.slotsFilled *
         _payoutAmount(requestId, requestExpiry(requestId));
@@ -468,7 +468,7 @@ contract Marketplace is SlotReservations, Proofs, StateRetrieval, Endian {
   }
 
   /**
-   * @notice Calculates the amount that should be payed out to a host that successfully finished the request
+   * @notice Calculates the amount that should be paid out to a host that successfully finished the request
    * @param requestId RequestId of the request used to calculate the payout
    * amount.
    * @param startingTimestamp timestamp indicating when a host filled a slot and
@@ -486,7 +486,7 @@ contract Marketplace is SlotReservations, Proofs, StateRetrieval, Endian {
       );
   }
 
-  /// @notice Calculates the amount that should be payed out to a host
+  /// @notice Calculates the amount that should be paid out to a host
   function _payoutAmount(
     RequestId requestId,
     uint256 startingTimestamp,
--- a/contracts/Marketplace.sol
+++ b/contracts/Marketplace.sol
@@ -350,9 +350,9 @@ contract Marketplace is SlotReservations, Proofs, StateRetrieval, Endian {
     _removeFromMySlots(slot.host, slotId);
 
     uint256 payoutAmount = _payoutAmount(
-        requestId,
-        slot.filledAt,
-        requestExpiry(requestId)
+      requestId,
+      slot.filledAt,
+      requestExpiry(requestId)
     );
     uint256 collateralAmount = slot.currentCollateral;
     _marketplaceTotals.sent += (payoutAmount + collateralAmount);
--- a/contracts/Marketplace.sol
+++ b/contracts/Marketplace.sol
@@ -29,9 +29,10 @@ contract Marketplace is SlotReservations, Proofs, StateRetrieval, Endian {
   struct RequestContext {
     RequestState state;
     uint256 slotsFilled;
-    /// @notice Tracks how much funds should be returned when Request expires to the Request creator
-    /// @dev The sum is deducted every time a host fills a Slot by precalculated amount that he should receive if the Request expires
-    uint256 expiryFundsWithdraw;
+    /// @notice Tracks how much funds should be returned to the client as not all funds might be used for hosting the request
+    /// @dev The sum starts on the full reward amount for the request and is deducted every time a host fills a Slot by precalculated amount that the host should receive
+    ///  if the he successfully finished the request (eq. for the time between when he filled the slot and request's end).
+    uint256 fundsToReturnToClient;
     uint256 startedAt;
     uint256 endsAt;
     uint256 expiresAt;
@@ -41,7 +42,7 @@ contract Marketplace is SlotReservations, Proofs, StateRetrieval, Endian {
     SlotState state;
     RequestId requestId;
     /// @notice Timestamp that signals when slot was filled
-    /// @dev Used for partial payouts when Requests expires and Hosts are paid out only the time they host the content.
+    /// @dev Used for calculating payouts as Hosts are payed based on time they actually host the content
     uint256 filledAt;
     uint256 slotIndex;
     /// @notice Tracks the current amount of host's collateral that is to be payed out at the end of Slot's lifespan.
@@ -112,8 +113,8 @@ contract Marketplace is SlotReservations, Proofs, StateRetrieval, Endian {
 
     _addToMyRequests(request.client, id);
 
-    uint256 amount = request.price();
-    _requestContexts[id].expiryFundsWithdraw = amount;
+    uint256 amount = request.maxPrice();
+    _requestContexts[id].fundsToReturnToClient = amount;
     _marketplaceTotals.received += amount;
     _transferFrom(msg.sender, amount);
 
@@ -142,6 +143,7 @@ contract Marketplace is SlotReservations, Proofs, StateRetrieval, Endian {
     Slot storage slot = _slots[slotId];
     slot.requestId = requestId;
     slot.slotIndex = slotIndex;
+    RequestContext storage context = _requestContexts[requestId];
 
     require(slotState(slotId) == SlotState.Free, "Slot is not free");
 
@@ -151,12 +153,9 @@ contract Marketplace is SlotReservations, Proofs, StateRetrieval, Endian {
     slot.host = msg.sender;
     slot.state = SlotState.Filled;
     slot.filledAt = block.timestamp;
-    RequestContext storage context = _requestContexts[requestId];
+
     context.slotsFilled += 1;
-    context.expiryFundsWithdraw -= _expiryPayoutAmount(
-      requestId,
-      block.timestamp
-    );
+    context.fundsToReturnToClient -= _payoutAmount(requestId, slot.filledAt);
 
     // Collect collateral
     uint256 collateralAmount = request.ask.collateral;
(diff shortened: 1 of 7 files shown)


test/Marketplace.test.js
```diff
@@ -652,14 +697,41 @@ describe("Marketplace", function () {
...
+        request.ask.reward
+
+      switchAccount(client)
+      await marketplace.withdrawFunds(slot.request)
+      const endBalance = await token.balanceOf(client.address)
+      expect(
+        maxPrice(request) -
+          (ACCOUNT_STARTING_BALANCE - endBalance) -
+          (request.expiry / 2) * request.ask.reward - // we have moved one slot to mid-expiry window
+          17 * request.ask.reward // There was 17 transaction for filling/advancing etc
```
emizzle, line 724: If you took a `startBalance` sample before calling `withdrawFunds`, shouldn't `expectedHostPayout` be equal to `(endBalance - startBalance)`? The idea being that if all four slots were filled for the entire duration, the client should have no refunds, and `(endBalance - startBalance)` would be 0. However, since one slot is freed after starting, and we expect a refund, the difference should be equal to the refund.

test/Marketplace.test.js
```diff
@@ -652,14 +697,41 @@ describe("Marketplace", function () {
...
+      const id = slotId(slot)
+      const filledAt = await currentTime()
+
+      await expect(marketplace.freeSlot(id)).to.emit(marketplace, "SlotFreed")
+      await waitUntilFinished(marketplace, requestId(request))
+
+      const expectedHostPayout =
+        ((await marketplace.requestEnd(requestId(request))).toNumber() -
+          filledAt) *
+        request.ask.reward
```
emizzle, line 715: Should this use `payoutForDuration`?

test/marketplace.js
```diff
@@ -14,11 +14,18 @@ async function waitUntilCancelled(request) {
...
+  request,
+  proof,
+  token,
+  ignoreSlotIndexes = []
+) {
+  await token.approve(contract.address, maxPrice(request) * request.ask.slots)
 
   for (let i = 0; i < request.ask.slots; i++) {
-    await contract.fillSlot(requestId(request), i, proof)
+    if (!ignoreSlotIndexes.includes(i))
```
emizzle, line 27: How can the contract be started if not all slots are filled? Maybe this needs its own function?

test/Marketplace.test.js
```diff
@@ -464,20 +464,30 @@ describe("Marketplace", function () {
...
+      await token.approve(marketplace.address, maxPrice(request))
       await marketplace.requestStorage(request)
       switchAccount(host)
       await token.approve(marketplace.address, request.ask.collateral)
     })
 
-    it("pays the host when contract has finished and returns collateral", async function () {
-      await waitUntilStarted(marketplace, request, proof, token)
+    it("pays the host when contract has finished for time he hosted the request and returns collateral", async function () {
+      await advanceTimeForNextBlock(request.expiry / 2)
```
emizzle, line 474: It's not clear to me why we need to advance the time

test/Marketplace.test.js
emizzle: Hmm actually I think it would not be 0 if there were no refunds as it depends when the slots were filled, and each transaction produces a block that is advanced by 1 second. So maybe instead of having a function `waitUntilStarted` with a parameter to not start certain slots, maybe there could be a function that fills certain slots and returns an array of start times? Those start times could be used to calculate how much each slot should be paid.

test/Marketplace.test.js
```diff
@@ -623,13 +644,37 @@ describe("Marketplace", function () {
...
-    it("withdraws to the client", async function () {
+    it("withdraw rest of funds for finished requests", async function () {
+      await waitUntilStarted(marketplace, request, proof, token)
+      await waitUntilFinished(marketplace, requestId(request))
+
+      switchAccount(client)
+      const startBalance = await token.balanceOf(client.address)
+      await marketplace.withdrawFunds(slot.request)
+      const endBalance = await token.balanceOf(client.address)
+      expect(endBalance - startBalance).to.lt(
```
emizzle, line 655:
```suggestion
      expect(endBalance - startBalance).to.be.lt(
```

test/Marketplace.test.js
```diff
@@ -623,13 +644,37 @@ describe("Marketplace", function () {
...
+        request.expiry * request.ask.reward
+      )
+    })
+
+    it("withdraws to the client when request is cancelled", async function () {
```
emizzle, line 660: Should there also be a test checking that cancelled requests pay out for slots that were filled before cancellation?

emizzle (approved):
Nice work! Overall a few suggestions and questions, but nothing blocking, so will approve now.

test/marketplace.js
emizzle: Right, but the contract won't be started if the optional parameter is used to exclude a slot from being filled...

AuHau: Well yes, but then you should know what you are doing if you are specifying that optional parameter...

0x-r4bbit:
I'd be great if we could land #162 first, so we can sure these changes pass the rules (or maybe rules need to be changed).

AuHau:
> I'd be great if we could land #162 first, so we can sure these changes pass the rules (or maybe rules need to be changed).

For sure! I am waiting for Mark's review anyhow ;-)

test/Marketplace.test.js
```diff
@@ -778,17 +842,48 @@ describe("Marketplace", function () {
       )
       const endBalance = await token.balanceOf(clientWithdrawRecipient.address)
       expect(endBalance - ACCOUNT_STARTING_BALANCE).to.equal(
-        price(request) - expectedPartialhostRewardRecipient
+        maxPrice(request) - expectedPartialhostRewardRecipient
       )
     })
+
+    it.skip("when slot is freed and not filled, client will get refunded the freed slot's funds", async function () {
```
markspanbroek, line 849: 🔴 It's not entirely clear to me what requirement you're trying to test here. I'd expect the requirement to be that the client is given the funds that remain after the hosts are paid. So if you know the payout to the hosts, then the rest is for the client. I'd then also structure the test in that way:
1. fill a slot halfway through the expiry window
2. when contract is finished, payout the hosts
3. withdraw funds, and check that it equals the paid amount minus the payouts to the hosts

test/Marketplace.test.js
```diff
@@ -490,52 +490,53 @@ describe("Marketplace", function () {
...
-      )
-    })
+    it("pays the host when contract has finished for time he hosted the request and returns collateral to host collateral address", async function () {
+      await advanceTimeForNextBlock(request.expiry / 2)
+      await marketplace.fillSlot(requestId(request), slot.index, proof)
+      const startTime = await currentTime()
 
-    it("pays to host reward address when contract has finished and returns collateral to host collateral address", async function () {
-      await waitUntilStarted(marketplace, request, proof, token)
+      await waitUntilStarted(marketplace, request, proof, token, [slot.index])
```
markspanbroek, line 504: 🟠 At a cursory glance, I'd expect this statement to wait until the slot with `slot.index` is started. Instead it does the opposite. You can solve this by making it a named parameter:

```suggestion
      await waitUntilStarted(marketplace, request, proof, token, { ignoreSlotIndexes: [slot.index] })
```

test/marketplace.js
markspanbroek: 🟠 I find this confusing as well. Why not add a `waitUntilSlotsFilled(indices)`? You can even use it to implement `waitUntilStarted()`

markspanbroek (commented):
Thanks for taking a stab at this @AuHau. I left a few comments.

0x-r4bbit:
Prover seems to be unhappy here as well btw. (yes the verify task still reports success, we need to get this fixed)

contracts/Marketplace.sol
```diff
@@ -373,24 +376,43 @@ contract Marketplace is Proofs, StateRetrieval, Endian {
...
+      // `fundsToReturnToClient` currently tracks funds to be returned for requests that successfully finish.
+      // When requests are cancelled, funds earmarked for payment for the duration
+      // between request expiry and request end (for every slot that was filled), should be returned to the client.
+      // Update `fundsToReturnToClient` to reflect this.
+      context.fundsToReturnToClient +=
+        context.slotsFilled *
+        _payoutAmount(requestId, requestExpiry(requestId));
+    } else if (state == RequestState.Failed) {
+      // For Failed requests the client is refunded whole amount.
+      context.fundsToReturnToClient = request.maxPrice();
```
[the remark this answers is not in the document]
AuHau: Yeah makes sense. Although I feel like it is a fairly weak mitigation. This mainly could open attacks on the hosts before ending of the request to wiggle out of payment for them... I guess since we are gonna have limited duration for Requests this should be OK as the cost of the attack would be most probably higher than what you would wiggled out from. But maybe the request's duration limit should stay on the lower end "for ever"?

contracts/Marketplace.sol
```diff
@@ -373,24 +376,43 @@ contract Marketplace is Proofs, StateRetrieval, Endian {
...
+
+    if (state == RequestState.Cancelled) {
+      context.state = RequestState.Cancelled;
+      emit RequestCancelled(requestId);
+
+      // `fundsToReturnToClient` currently tracks funds to be returned for requests that successfully finish.
+      // When requests are cancelled, funds earmarked for payment for the duration
+      // between request expiry and request end (for every slot that was filled), should be returned to the client.
+      // Update `fundsToReturnToClient` to reflect this.
+      context.fundsToReturnToClient +=
```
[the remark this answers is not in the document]
AuHau: > Doesn't this mean that the amount of funds that a client receives depends on when it calls withdrawFunds().

No, it does not depend on that. You are most probably referencing the `context.slotsFilled` right? That variable is not lowered when calling `freeSlot` in states **Failed/Cancelled/Finished**, only when the node would call that function during "running" of the Request, which again would not effect this function as it can be called only in those states...

contracts/Marketplace.sol
```diff
@@ -373,24 +376,43 @@ contract Marketplace is Proofs, StateRetrieval, Endian {
...
     RequestContext storage context = _requestContexts[requestId];
-    require(context.state == RequestState.New, "Invalid state");
+    RequestState state = requestState(requestId);
+    require(
+      state == RequestState.Cancelled ||
+        state == RequestState.Failed ||
+        state == RequestState.Finished,
+      "Invalid state"
+    );
+    require(context.fundsToReturnToClient != 0, "Nothing to withdraw");
```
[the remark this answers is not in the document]
AuHau: But a good idea is that I should have a test that will verify the double-spend protection...

contracts/Marketplace.sol
```diff
@@ -373,24 +376,43 @@ contract Marketplace is Proofs, StateRetrieval, Endian {
...
 
-    emit RequestCancelled(requestId);
-
-    uint256 amount = context.expiryFundsWithdraw;
+    uint256 amount = context.fundsToReturnToClient;
     _marketplaceTotals.sent += amount;
     assert(_token.transfer(withdrawRecipient, amount));
+
+    // We zero out the funds tracking in order to prevent double-spends
+    context.fundsToReturnToClient = 0;
```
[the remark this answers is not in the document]
AuHau: I don't think it is necessary. Also the reason why I would avoid this approach is because we will lose information about how the request terminated if we move from the current terminating states into "one that catches them all".

test/Marketplace.test.js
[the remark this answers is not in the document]
AuHau: Haha yeah, because it does not work 🤪 I wanted to return to this test to rewrite it properly.

AuHau: It needs rework agree, but I am trying to be here a bit more explicit in my assertions here, because doing assertions on "something is something minus something" is IMO quite week, because if there is bug in the calculation of the reward, then this would pass and it would not catch the sizing of the propotions...

Also, I am trying to test case where request does finish with not all slots filled and that the reward for that slot is returned.

contracts/Marketplace.sol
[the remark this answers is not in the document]
markspanbroek: No, I didn't mean `context.slotsFilled`. I was thinking about the additions to `context.fundsToReturnToClient` in `freeSlot`. But looking through the code again, I realize that this only happens in `_forciblyFreeSlot()`, and that can only happen while the slot is Filled.

contracts/Marketplace.sol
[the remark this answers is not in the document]
markspanbroek: Good point. It's not a very practical attack anyway.

I would not be in favor of keeping the duration limit low for ever. It limits the usefulness of the network.

test/Marketplace.test.js
markspanbroek: Alright, I'll look at it again after you rework the test.

contracts/Marketplace.sol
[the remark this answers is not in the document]
markspanbroek: Good  point, we still need the distinction on how a request ended to handle `freeSlot()` correctly.

contracts/Marketplace.sol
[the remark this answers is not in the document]
markspanbroek: You could consider adding a boolean `context.clientPaid` to make it more explicit.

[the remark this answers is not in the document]
AuHau: IMHO this mechanism is sufficient if properly documented, but at the same time I don't have strong opinions about it either, so if you think I should go with the flag approach, then let me know.

contracts/Marketplace.sol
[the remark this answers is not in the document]
AuHau: > I would not be in favor of keeping the duration limit low for ever. It limits the usefulness of the network.

I disagree about the limiting of the usefulness of the network. IMHO we can nicely plug in account abstraction for "recurring request renewal" to host requests "for ever". If we allow long requests in the system, then we will open doors to this attack that you mentioned before.

contracts/Marketplace.sol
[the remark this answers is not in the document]
emizzle: I agree that a flag would be more explicit, albeit it does incur a higher gas cost for storing additional state.

contracts/Marketplace.sol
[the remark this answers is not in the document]
emizzle: > If we allow long requests in the system, then we will open doors to this attack that you mentioned before.

30 days is long enough for this attack. A DDOS attack would only need to be long enough that it can force slots to be freed right near the end of the contract. The needed duration of the attack would depend on the proof frequency, however 30 days should be way more time than is needed.

I really like the idea of burning a percentage of the funds that would be returned to the client, however:
1. I see the cost of this attack as the mitigation needed, as there is no financial gain to be had from performing the attack. The only gain is disruption of the network. 
2. From an honest user's perspective, getting all funds returned in case the network fails to live up to its promises is a minimal level of consolation needed. If the user were to lose funds when the network fails (burned), I'm not sure how eager they'd be to come back and use Codex again.

test/Marketplace.test.js
emizzle: Makes sense. Maybe add a comment?

contracts/Marketplace.sol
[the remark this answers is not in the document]
AuHau: > 30 days is long enough for this attack. A DDOS attack would only need to be long enough that it can force slots to be freed right near the end of the contract.

I am talking here about the client's attack, which could "wiggle out" of the payment for the stored data if he DDOSed the providers at the end of the storage requests and hence got returned the full reward amount.

What I am mentioning here is that if the limit of storage request duration is low, then the price for the request is low, which I would argue that the gain from such an attack should be small enough to justify a DDOS attack on several node providers. 

ChatGPT (hence take it with a grain of salt) says the cheapest DDOS starts at 10$ per hour. Hence, you would need at least 24 hours for several days in order for the providers to get kicked out of the request and make it fail. So the cost to kick out a single provider starts at least on: `$10 * 24 * 4 = $960` (the 4 is the arbitrary number for how many slashes it takes before the node is kicked out). Based on the tolerance setting, there might need to be more providers kicked out as well.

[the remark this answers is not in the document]
emizzle: I'm pretty sure we're saying the same thing 😅  I was only trying to point out that our "low duration" of 30 days is already long enough to allow for DDOS attacks and so this is somewhat of a moot point in arguing for keeping contract duration low.

> the gain from such an attack
Technically, there is no gain except "free" storage and network disruption, as there is no flow of *additional* funds to the client. They would simply get their own funds returned to them. Well they might get reimbursed for the gas cost associated with creating the request, but there's no way that would exceed the cost of a DDOS attack. It'll almost definitely be cheaper to actually pay for the storage 😄

test/Marketplace.test.js
AuHau: Yeah, this is something I more or less did. Please check the reworked test.

test/Marketplace.test.js
AuHau: Done, please check again.

AuHau:
I have implemented your feedback and reworked the marketplace tooling in tests where `waitUntilStarted()` now returns the array of expected payouts for each filled slot and I use that to make the calculations - IMHO this looks much better and I would say is good to go.

test/Marketplace.test.js
```diff
@@ -739,7 +785,36 @@ describe("Marketplace", function () {
...
+      const endBalancePayout = await token.balanceOf(
+        clientWithdrawRecipient.address
+      )
+
+      expect(endBalanceClient).to.equal(startBalanceClient)
+      // As all the request's slots will get filled and request will start and successfully finishes,
+      // then the upper bound to how much the client gets returned is the cumulative reward for all the
+      // slots for expiry window. This limit is "inclusive" because it is possible that all slots are filled
+      // at the time of expiry and hence the user would get the full "expiry window" reward back.
+      expect(endBalancePayout - startBalancePayout).to.be.lte(
```
markspanbroek, line 812: Perhaps also a check here that the payout is > 0?

test/marketplace.js
```diff
@@ -14,12 +14,38 @@ async function waitUntilCancelled(request) {
...
+      request,
+      await currentTime(),
+      requestEnd
+    )
   }
+
+  return payouts
+}
+
+function slotsArrayExcept(request, ignoredIndex) {
```
markspanbroek, line 35: This is not used anywhere, as far as I can tell.

markspanbroek (approved):
Thanks @AuHau! The tests are very nice now.

Issue #157, opened 2024-08-15
`fillSlot` should guard against not allowed state transitions


AuHau:
`fillSlot` should verify that the Request is in states `New` or `Started`, if we would allow for the rest of the states then we could get illegal state transitions from `Cancelled`/`Failed`/`Finished` to `Started`.

Credit to @0x-r4bbit as he found this as part of Certora work.

0x-r4bbit:
Working on that atm.

0x-r4bbit:
Quick update on this, we have a fix for this in place but yet have to fix some rules that are failing after we've landed #158

0x-r4bbit:
This will be fixed with #161

AuHau:
In the end it seems like this line actually prevent the behavior described in this issue:
https://github.com/codex-storage/codex-contracts-eth/blob/8bbb4257f7b42bf257303d4f504674adbebdefee/contracts/Marketplace.sol#L139

0x-r4bbit:
Closing this as discussed above.

0x-r4bbit:
Just to add this here for reference:

The reason this issue came up in the first place, is because we had a counter example in one of the rules that has shown that an already `Finished` request can be changed to `Started`, which is actually forbidden.

The reason it could do that was because the prover starts with a finished request and calls `fillSlot()` with a slotId that points to a request with id `0`. In practice, id `0` is not really possible, but the prover doesn't know that.

When `requestId == 0` then this ultimately causes the code to think the `slotstate` of the slot in question is `Free` (which is not necessarily the case).

Long story short, the prover shouldn't call the function with `requestId = 0` but we also can't blindly require that the `requestId != 0` because there's other cases that would be excluded from the rule.

For more context see: https://github.com/codex-storage/codex-contracts-eth/pull/162#discussion_r1728772997

PR #162, opened 2024-08-21
Make certora rules pass again

### LINKED ISSUE #161: Make certora rules pass again
After we've corrected our certora config in #158 it uncovered that many of our rules were actually failing.
We should get those passing again to make sure we catch possible bugs in the future.

0x-r4bbit:
We've had a bug in our certora configuration which resulted in may of our rules to pass, which are actually failing, now that the configuration has been updated.

This PR introduces a bunch of commits that in aggregate fix the broken rules again.

Closes #161

### DIFF
--- a/certora/specs/Marketplace.spec
+++ b/certora/specs/Marketplace.spec
@@ -170,6 +170,18 @@ invariant finishedSlotAlwaysHasFinishedRequest(env e, Marketplace.SlotId slotId)
     currentContract.slotState(e, slotId) == Marketplace.SlotState.Finished =>
         currentContract.requestState(e, currentContract.slots(e, slotId).requestId) == Marketplace.RequestState.Finished;
 
+// STATUS - verified
+// paid slot always has finished or cancelled request
+// https://prover.certora.com/output/6199/6217a927ff2c43bea1124f6ae54a78fb?anonymousKey=d5e09d0d12658fd6b5f298fc04cea88da892c62d
+invariant paidSlotAlwaysHasFinishedOrCancelledRequest(env e, Marketplace.SlotId slotId)
+     currentContract.slotState(e, slotId) == Marketplace.SlotState.Paid =>
+         currentContract.requestState(e, currentContract.slots(e, slotId).requestId) == Marketplace.RequestState.Finished || currentContract.requestState(e, currentContract.slots(e, slotId).requestId) == Marketplace.RequestState.Cancelled
+    { preserved {
+        // ensures we start with a paid slot
+        require currentContract.slotState(e, slotId) == Marketplace.SlotState.Paid;
+      }
+    }
+
 /*--------------------------------------------
 |                 Properties                 |
 --------------------------------------------*/
--- a/certora/specs/Marketplace.spec
+++ b/certora/specs/Marketplace.spec
@@ -136,7 +136,7 @@ invariant totalSupplyIsSumOfBalances()
     to_mathint(Token.totalSupply()) == sumOfBalances;
 
 invariant requestStartedWhenSlotsFilled(env e, Marketplace.RequestId requestId, Marketplace.SlotId slotId)
-    to_mathint(currentContract.requestContext(e, requestId).slotsFilled) == to_mathint(currentContract.getRequest(e, requestId).ask.slots) => currentContract.requestState(e, requestId) == Marketplace.RequestState.Started;
+    currentContract.requestState(e, requestId) == Marketplace.RequestState.Started => to_mathint(currentContract.getRequest(e, requestId).ask.slots) - to_mathint(currentContract.requestContext(e, requestId).slotsFilled) <= to_mathint(currentContract.getRequest(e, requestId).ask.maxSlotLoss);
 
 // STATUS - verified
 // https://prover.certora.com/output/6199/6e2383ea040347eabeeb1008bc257ae6?anonymousKey=e1a6a00310a44ed264b1f98b03fa29273e68fca9
--- a/certora/specs/Marketplace.spec
+++ b/certora/specs/Marketplace.spec
@@ -163,6 +163,13 @@ invariant failedRequestAlwaysEnded(env e, Marketplace.RequestId requestId)
     currentContract.requestState(e, requestId) == Marketplace.RequestState.Failed =>
         currentContract.requestContext(e, requestId).endsAt < lastBlockTimestampGhost;
 
+// STATUS - verified
+// finished slot always has finished request
+// https://prover.certora.com/output/6199/3371ee4f80354ac9b05b1c84c53b6154?anonymousKey=eab83785acb61ccd31ed0c9d5a2e9e2b24099156
+invariant finishedSlotAlwaysHasFinishedRequest(env e, Marketplace.SlotId slotId)
+    currentContract.slotState(e, slotId) == Marketplace.SlotState.Finished =>
+        currentContract.requestState(e, currentContract.slots(e, slotId).requestId) == Marketplace.RequestState.Finished;
+
 /*--------------------------------------------
 |                 Properties                 |
 --------------------------------------------*/
--- a/certora/specs/Marketplace.spec
+++ b/certora/specs/Marketplace.spec
@@ -120,12 +120,14 @@ function canStartRequest(method f) returns bool {
 }
 
 function canFinishRequest(method f) returns bool {
-    return f.selector == sig:freeSlot(Marketplace.SlotId, address, address).selector;
+    return f.selector == sig:freeSlot(Marketplace.SlotId, address, address).selector ||
+        f.selector == sig:freeSlot(Marketplace.SlotId).selector;
 }
 
 function canFailRequest(method f) returns bool {
     return f.selector == sig:markProofAsMissing(Marketplace.SlotId, Periods.Period).selector ||
-        f.selector == sig:freeSlot(Marketplace.SlotId, address, address).selector;
+        f.selector == sig:freeSlot(Marketplace.SlotId, address, address).selector ||
+        f.selector == sig:freeSlot(Marketplace.SlotId).selector;
 }
 
 /*--------------------------------------------
@@ -261,6 +263,9 @@ rule functionsCausingRequestStateChanges(env e, method f) {
 
     // RequestState.New -> RequestState.Started
     assert requestStateBefore == Marketplace.RequestState.New && requestStateAfter == Marketplace.RequestState.Started => canStartRequest(f);
+
+    // RequestState.New -> RequestState.Cancelled
+    assert requestStateBefore == Marketplace.RequestState.New && requestStateAfter == Marketplace.RequestState.Cancelled => canCancelRequest(f);
 
     // RequestState.Started -> RequestState.Finished
     assert requestStateBefore == Marketplace.RequestState.Started && requestStateAfter == Marketplace.RequestState.Finished => canFinishRequest(f);
--- a/certora/specs/Marketplace.spec
+++ b/certora/specs/Marketplace.spec
@@ -222,6 +222,21 @@ rule totalSentCannotDecrease(env e, method f) {
     assert total_after >= total_before;
 }
 
+rule slotIsFailedOrFreeIfRequestHasFailed(env e, method f) {
+    calldataarg args;
+    Marketplace.SlotId slotId;
+
+    requireInvariant paidSlotAlwaysHasFinishedOrCancelledRequest(e, slotId);
+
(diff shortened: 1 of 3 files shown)


0x-r4bbit:
This is a WIP, will mark as ready for review once ready.

certora/specs/Marketplace.spec
```diff
@@ -138,6 +153,52 @@ invariant slotMissedShouldBeEqualToNumberOfMissedPeriods(env e, Marketplace.Slot
...
+
+// STATUS - verified
+// cancelled slot always belongs to cancelled request
+// https://prover.certora.com/output/6199/80d5dc73d406436db166071e277283f1?anonymousKey=d5d175960dc40f72e22ba8e31c6904a488277e57
+invariant cancelledSlotAlwaysHasCancelledRequest(env e, Marketplace.SlotId slotId)
+    currentContract.slotState(e, slotId) == Marketplace.SlotState.Cancelled =>
+        currentContract.requestState(e, currentContract.slots(e, slotId).requestId) == Marketplace.RequestState.Cancelled;
+
+// STATUS - verified
+// requests always ask for slot amount greaater than 0
```
AuHau, line 197:
```suggestion
// requests always ask for slot amount greater than 0
```

certora/specs/Marketplace.spec
```diff
@@ -138,6 +153,52 @@ invariant slotMissedShouldBeEqualToNumberOfMissedPeriods(env e, Marketplace.Slot
...
+// https://prover.certora.com/output/6199/066a1cf8f74a457eaa820263d30f66be?anonymousKey=2217c966df8a9c32fe6112cd21fc0a8df601149f
+invariant requestAskSlotsGreaterThanZero(env e, Marketplace.RequestId requestId)
```
AuHau, line 199: Interesting, we are actually not validating that `request.ask.slots` is > 0, maybe we should add that. 
Does invariant "test" this? Or just constrains the prover to not test values <= 0?

contracts/Marketplace.sol
```diff
@@ -408,6 +408,14 @@ contract Marketplace is Proofs, StateRetrieval, Endian {
     _;
   }
 
+  modifier requestNotFinished(RequestId requestId) {
+    require(
+      _requestContexts[requestId].state != RequestState.Finished,
+      "Request is cancelled or finished"
+    );
```
AuHau, line 415: Nice! But, I don't think this is enough. You should use instead `requestState(requestId)` which checks dynamically on the request's state, which is not always "propagated" into the `requestContext.state`variable. For example `Cancelled` state is saved to `requestContext.state` only once the Client calls `withdrawFunds`. Also `Finished` != `Cancelled`... 

I am curious, if the prover is passing with this fix, maybe there is some over constraining happening in your rules? Because I would have expected that it should find counter-examples for this, no?

AuHau (changes requested):
Generally LGTM, but I believe we need to change the `requestNotFinished` modifier.

certora/specs/Marketplace.spec
```diff
@@ -71,11 +71,22 @@ hook Sload bool defaultValue _missing[KEY MarketplaceHarness.SlotId slotId][KEY
 
 hook Sstore _missing[KEY MarketplaceHarness.SlotId slotId][KEY Periods.Period period] bool defaultValue {
     _missingMirror[slotId][period] = defaultValue;
-    _missedCalculated[slotId] = _missedCalculated[slotId] + 1;
+    if (defaultValue) {
+        _missedCalculated[slotId] = _missedCalculated[slotId] + 1;
+    } else {
+        _missedCalculated[slotId] = _missedCalculated[slotId] - 1;
+    }
```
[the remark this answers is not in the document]
0x-r4bbit: So we have this `missedCalculated[slotId]` ghost, which aims to keep track of how many times a slot has been marked as missed.

We're hooking into `_missing` map `SSTORE` operations. In other words, whenever a slot period is `missing` is set `true` we increase the counter. Otherwise we decrease it.

I don't actually think there's a case currently in code where it's set `false` so it should only increase atm.

certora/specs/Marketplace.spec
0x-r4bbit: The invariant proofs that, for any given request, if the `ask.slots` are `> 0` they stay `> 0`.

It doesn't actually proof the case you've described. I'll provide a fix and try to write a rule that proofs it as well.
Good catch!

contracts/Marketplace.sol
0x-r4bbit: Very good catch. I remember actually using `requestState()` while debugging something and then I rolled back to this.

Will verify with `requestState()` again and update this commit. Thank you for pointing this out!

0x-r4bbit: @AuHau passes with `requestState()` 

https://prover.certora.com/output/6199/64289db2b3044f21937c1955f1bf695a?anonymousKey=35a74f14ec6be98fe9fab29f03187cfd5c810178

certora/specs/Marketplace.spec
0x-r4bbit: > The invariant proofs that, for any given request, if the ask.slots are > 0 they stay > 0.

Also just realized that this invariant isn't very meaningful. Gonna remove it and add a test for `ask.slots > 0` check anyways.

certora/specs/Marketplace.spec
```diff
@@ -138,6 +153,52 @@ invariant slotMissedShouldBeEqualToNumberOfMissedPeriods(env e, Marketplace.Slot
...
+invariant failedRequestAlwaysEnded(env e, Marketplace.RequestId requestId)
+    currentContract.requestState(e, requestId) == Marketplace.RequestState.Failed =>
+        currentContract.requestContext(e, requestId).endsAt < lastBlockTimestampGhost;
+
+// STATUS - verified
+// finished slot always has finished request
+// https://prover.certora.com/output/6199/3371ee4f80354ac9b05b1c84c53b6154?anonymousKey=eab83785acb61ccd31ed0c9d5a2e9e2b24099156
+invariant finishedSlotAlwaysHasFinishedRequest(env e, Marketplace.SlotId slotId)
+    currentContract.slotState(e, slotId) == Marketplace.SlotState.Finished =>
+        currentContract.requestState(e, currentContract.slots(e, slotId).requestId) == Marketplace.RequestState.Finished;
```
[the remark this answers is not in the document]
0x-r4bbit: Thank you! I'll look into this

certora/specs/Marketplace.spec
0x-r4bbit: @AuHau added a separate PR for this fix here: https://github.com/codex-storage/codex-contracts-eth/pull/163

certora/specs/Marketplace.spec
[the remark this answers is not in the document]
AuHau: > I don't actually think there's a case currently in code where it's set false so it should only increase atm.

Fair, I just don't understand how you want to employ the decreasing logic in some potential tests? It seems to me like a construct that could lead to some unexpected behavior or potentially hiding some bugs? I would have expected tests like "did proof got missed" or similar using assertions on this array by expecting the slot's counter to increase. If there is option that it could also decrease, then you could lend in a state which would have missed proof but it would not be detected...

contracts/Marketplace.sol
```diff
@@ -408,6 +408,14 @@ contract Marketplace is Proofs, StateRetrieval, Endian {
     _;
   }
 
+  modifier requestNotFinished(RequestId requestId) {
```
AuHau, line 411: Soooo, I was checking something around the `fillSlot` and actually I noticed that there is the line `require(slotState(slotId) == SlotState.Free, "Slot is not free");` which inside actually checks for `requestState()` and if the request is cancelled/failed/finished, then it will returns something else then `SlotState.Free`, soooo actually this issue is not relevant? 😅
https://github.com/codex-storage/codex-contracts-eth/issues/157

AuHau: Can you also please double-check this @0x-r4bbit if I see this correctly? 

https://github.com/codex-storage/codex-contracts-eth/blob/8bbb4257f7b42bf257303d4f504674adbebdefee/contracts/Marketplace.sol#L139

If so, let's then remove this modifier? I guess I would like to find maybe some better way how to communicate this in the code as it is not clear, but that is refactor for some other day.

0x-r4bbit: Oh well...

So I've removed the modifier and ran the prover on `allowedRequestStateChanges` rule (https://prover.certora.com/output/6199/c0f500e68f0441a7b7f371aa6e9cba12?anonymousKey=2014a4af0ac07d3847d0ae7b200ef07510dab545) and `functionsCausingRequestStateChanges` rule (https://prover.certora.com/output/6199/f19c47de699e4d9694f108b7693e75c5?anonymousKey=b5dccf783024345142761adb6205426ba6cfe14f) and they are both passing.

I can't remember what scenario caused a counter example where the modifier was needed.
But it looks like it no longer is.

Thanks once again for your using your eagle eyes!

certora/specs/Marketplace.spec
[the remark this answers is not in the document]
0x-r4bbit: I'm going to double check this and update the hook if necessary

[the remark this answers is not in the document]
0x-r4bbit: Invariant passes without decrease, gonna remove that.
Here's the prover run: https://prover.certora.com/output/6199/6a076ed7e7a149abaa17d412edc97163?anonymousKey=4eb37acdfecbef4ab032f070b5bc124730d86fad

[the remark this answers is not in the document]
0x-r4bbit: Updated the PR.

contracts/Marketplace.sol
AuHau: No worries, I also originally thought that the bug is valid ;-)

AuHau (approved):
LGTM, thanks!

test/Marketplace.test.js
```diff
@@ -293,14 +293,6 @@ describe("Marketplace", function () {
...
     })
 
-    it("is rejected when request is finished", async function () {
-      await waitUntilStarted(marketplace, request, proof, token)
-      await waitUntilFinished(marketplace, slot.request)
-      await expect(
-        marketplace.fillSlot(slot.request, slot.index, proof)
-      ).to.be.revertedWith("Slot is not free")
-    })
-
```
AuHau, line 303: Btw. why are you removing this test? IMHO it should stay here no? That is valid test-case...

0x-r4bbit: wtf. That shouldn't have happened. I've removed another one that I added earlier.
Will update this.

contracts/Marketplace.sol
0x-r4bbit: @AuHau Here you go: This is now failing with all rules enabled: https://prover.certora.com/output/532476/1955a12c799c4c749d4e7f2f94549607?anonymousKey=82e229207950662a9c624d7b0864bf40b1edee4e

0x-r4bbit: Ah, that's the issue again with the prover runnign `fillSlot()` with `requestId = 0` which will always give you a `Slot.Free` slot, even though the `RequestState` is `Finished`.

Gonna fix this.

0x-r4bbit: Alright, here's a prover run with a fix: https://prover.certora.com/output/6199/0ea1d80bc3ad42438c9303d2b3604b4f?anonymousKey=5faeef8e81a4b5ee0d30f77b5ee984f1c5d26410

Here an isolated failing one for comparison:
https://prover.certora.com/output/6199/81939b2b12d74a5cae5e84ceadb901c0?anonymousKey=a4ad6268598a1077ecfce75493b0c0f9bc3b17a0

test/Marketplace.test.js
0x-r4bbit: Undid this change. Sorry, this was a mistake.

contracts/Marketplace.sol
0x-r4bbit: Updated the PR with the fixes.

0x-r4bbit:
All rules seem to have passed https://prover.certora.com/output/532476/9cb7d1288d104cbeabee6ce2590115ca?anonymousKey=5cee79047c76e22e606ab276f65ccd102c48d167

certora/specs/Marketplace.spec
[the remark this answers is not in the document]
0x-r4bbit: Here's a PR aiming to address your comment https://github.com/codex-storage/codex-contracts-eth/pull/167
Let's continue the discussion there.

PR #163, opened 2024-08-23
fix(Marketplace): ensure requests include ask with sufficient slots

0x-r4bbit:
There is a missing check in `requestStorage()` on whether the `Request` contains an `Ask` where its `slots` is `> 0`.

This allows for making storage request without slots. Not harmful but not a valid state of the system either.

This commit adds that check and a test with batteries included.


AuHau (approved):
LGTM, thanks!

0x-r4bbit:
Rebased on top of latest master, waiting for CI to be green (and double check prover), will then merge.

Issue #127, opened 2024-07-30
Implement application property: The contract has enough funds to pay back everything


0x-r4bbit:
This issue is about implementing the application property that the marketplace owns enough funds to pay back its storage requesters and storage providers.

Exact details of this issue will be worked out with the smart contracts team and certora.

PR #169, opened 2024-08-27
[WIP] spec: The contract has enough funds to pay back everything

gravityblast:
Part of #127 

Add a spec to check that the contract balance increases when new storage requests are added.

certora run started from local command: 

https://prover.certora.com/output/27938/7c09c206648f4d8d9e26ed839d14f0b1?anonymousKey=40b24d364cac6b8df280b8b9e2bc62551c3bdfdc

### DIFF
--- a/certora/specs/Marketplace.spec
+++ b/certora/specs/Marketplace.spec
@@ -14,6 +14,14 @@ ghost mathint sumOfBalances {
     init_state axiom sumOfBalances == 0;
 }
 
+ghost mathint requestsCount {
+    init_state axiom requestsCount == 0;
+}
+
+ghost mathint sumOfAllRequestPrices {
+    init_state axiom sumOfAllRequestPrices == 0;
+}
+
 hook Sload uint256 balance Token._balances[KEY address addr] {
     require sumOfBalances >= to_mathint(balance);
 }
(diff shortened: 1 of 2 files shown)


certora/specs/Marketplace.spec
```diff
@@ -191,6 +228,9 @@ invariant cancelledSlotAlwaysHasCancelledRequest(env e, Marketplace.SlotId slotI
     currentContract.slotState(e, slotId) == Marketplace.SlotState.Cancelled =>
         currentContract.requestState(e, currentContract.slots(e, slotId).requestId) == Marketplace.RequestState.Cancelled;
 
+invariant requestsCountIsGreaterOrEqualToZero()
+    requestsCount >= 0;
```
0x-r4bbit, line 232: @gravityblast did we see counter examples where we had a negative request count?
Wondering otherwise how useful this invariant is.

gravityblast: @0x-r4bbit  yes we had a run with `requestsCount` starting with `-1` because the `init_state axiom` works only with invariants, so as discussed with the Certora team we need this invariant to make sure it's not a negative number in the rule

0x-r4bbit:
@gravityblast can you rebase this on top of latest `master`?
It has some fixes now and CI so properly report errors.

AuHau:
Is this ready for review or not? On one hand it is marked as ready for review but at the same time there is `[WIP]` in the title - I am confused 😅 I would suggest sticking to using the Github draft/ready PR's status to signal if PR is ready for review or not.

PR #170, opened 2024-08-27
chore(CI): update certora-cli to 7.10.2

0x-r4bbit:
This should fix the bug of certora-cli not properly completing with non zero exit codes on errors, which ultimately results in CI tasks to give false positives.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,7 +52,7 @@ jobs:
           java-package: 'jre'
 
       - name: Install Certora CLI
-        run: pip3 install certora-cli==7.10.1
+        run: pip3 install certora-cli==7.10.2
 
       - name: Install Solidity
         run: |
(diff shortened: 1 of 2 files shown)


AuHau (approved):

0x-r4bbit:
As we can see, this commit is (correctly) failing on CI. I'm going to cherry-pick #168 on top of this so we end up with a green CI again. All other pending PRs should be rebased on top of that.

0x-r4bbit:
@AuHau ocne this is green, we should be gtg (gonna take a while tho)

AuHau:
> As we can see, this commit is (correctly) failing on CI. I'm going to cherry-pick https://github.com/codex-storage/codex-contracts-eth/pull/168 on top of this so we end up with a green CI again. All other pending PRs should be rebased on top of that.

Ok, or lets just merge that and rebase this ;-)

Issue #165, opened 2024-08-26
Make use of ghosts instead of harness calls for requestContext and slots(id).requestId


0x-r4bbit:
There are things we've introduced harness functionality for to access fields that where otherwise unaccessible to us within CVL: requestContexts and `slots(slotId).requestId`.

Both were pointed out by Certora (one example is here: https://github.com/codex-storage/codex-contracts-eth/pull/162/files#diff-476cf79e9c5e1e4008e909b0fb5983bb4a5c59a364fad0d0bb80357bc7551c4cR173)

This issue is about revisiting the usages of `requestContext()` and `slots()` provided by `MarketplaceHarness` and ideally replace those with ghost variables, so we can reduce the harness surface.

PR #171, opened 2024-08-29
chore(certora): verify possible slot state transitions

gravityblast:
This PR adds a rule to check the possible slot transitions saved in storage and adss an helper function to check the consistency of the slot attributes.

### DIFF
--- a/certora/harness/MarketplaceHarness.sol
+++ b/certora/harness/MarketplaceHarness.sol
@@ -7,12 +7,22 @@ import {IGroth16Verifier} from "../../contracts/Groth16.sol";
 import {MarketplaceConfig} from "../../contracts/Configuration.sol";
 import {Marketplace} from "../../contracts/Marketplace.sol";
 import {RequestId, SlotId} from "../../contracts/Requests.sol";
+import {Requests} from "../../contracts/Requests.sol";
 
 contract MarketplaceHarness is Marketplace {
-    constructor(MarketplaceConfig memory config, IERC20 token, IGroth16Verifier verifier) Marketplace(config, token, verifier) {}
+    constructor(MarketplaceConfig memory config, IERC20 token, IGroth16Verifier verifier)
+        Marketplace(config, token, verifier)
+    {}
 
     function publicPeriodEnd(Period period) public view returns (uint256) {
         return _periodEnd(period);
     }
-}
 
+    function slots(SlotId slotId) public view returns (Slot memory) {
+        return _slots[slotId];
+    }
+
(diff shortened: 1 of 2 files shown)


certora/harness/MarketplaceHarness.sol
```diff
@@ -7,12 +7,22 @@ import {IGroth16Verifier} from "../../contracts/Groth16.sol";
...
+    {}
 
     function publicPeriodEnd(Period period) public view returns (uint256) {
         return _periodEnd(period);
     }
-}
 
+    function slots(SlotId slotId) public view returns (Slot memory) {
+        return _slots[slotId];
+    }
```
0x-r4bbit, line 23: We had this harness API before and then it was removed in #165 
I had a discussion with the Certora team about this. Specifically discussing, when it makes sense to favour ghost variables over functions like these to keep the harness surface low.

I don't think there's a good reason not to have these functions as long as you don't change the logic of the original contract, so I think this is fine.

Otherwise, you could introduced ghost variables for the fields you're interested in.

certora/specs/Marketplace.spec
```diff
@@ -341,6 +358,25 @@ rule functionsCausingRequestStateChanges(env e, method f) {
...
+    Marketplace.SlotState slotStateBefore = currentContract.slotState(e, slotId);
+    f(e, args);
+    Marketplace.SlotState slotStateAfter = currentContract.slotState(e, slotId);
+
+    // SlotState.Finished -> SlotState.Paid
+    assert slotStateBefore == Marketplace.SlotState.Finished && slotStateAfter == Marketplace.SlotState.Paid => canMakeSlotPaid(f);
+
+    // SlotState.Free -> SlotState.Filled
+    assert slotStateBefore != Marketplace.SlotState.Filled && slotStateAfter == Marketplace.SlotState.Filled => canFillSlot(f);
+    assert slotStateBefore != Marketplace.SlotState.Filled && slotStateAfter == Marketplace.SlotState.Filled => slotStateBefore == Marketplace.SlotState.Free;
```
0x-r4bbit, line 377: Do you think it makes sense to have an `allowedSlotStateChanges` rule as well? Right now it seems it's doing a mix of "allowedSlotStateChanges" and "functionsCausingSlotStateChanges".

Not something that has to be done now, but I see some similarities to the request state change rules.

0x-r4bbit: Also, what about any `SlotState.Cancelled` variations?

0x-r4bbit (approved):
This LGTM!
I've left some comments. No blockers, but worth exploring IMO

0x-r4bbit:
Please provide a prover run next time.

PR #176, opened 2024-09-18
feat(slot-reservations): Allow slots to be reserved

### LINKED ISSUE #175: feat(slot-reservations): Allow slots to be reserved
Allows reservation of slots, without an implementation of the expanding window.

- [ ] Add a function called `reserveSlot(address, SlotId)`, that allows three unique addresses per slot to be reserved, that returns bool if successful. 
  - Use `mapping(SlotId => EnumerableSet.AddressSet)` - Return false if the address could not be added to the set (if `EnumerableSet.add` returns false)
- [ ] Add `canReserveSlot(address, SlotId)` 
  - Return `true` if set of reservations is less than 3 and the set doesn't already contain the address 
  - Return `true` otherwise (for now, later add in logic for checking the address is inside the expanding window) 
  - Call `canReserveSlot` from `reserveSlot` as a `require` or invariant

emizzle:
Closes #175. 

These functions are not called yet. They are being implemented as part of a [longer list of tasks for slot reservations](https://github.com/codex-storage/codex-pm/issues/45) to prevent PRs from being too large.

Allows reservation of slots, without an implementation of the expanding window.

- Add a function called `reserveSlot(address, SlotId)`, that allows three unique addresses per slot to be reserved. 
  - Revert if reservations full or if provider address already reserved slot
  - Use `mapping(SlotId => EnumerableSet.AddressSet)`
- Add `canReserveSlot(address, SlotId)`
  - Return `true` if set of reservations is less than 3 and the set doesn't already contain the address
  - Return `true` otherwise (for now, later add in logic for checking the address is inside the expanding window)
  - Call `canReserveSlot` from `reserveSlot` as a `require` or invariant

### DIFF
--- a/contracts/Marketplace.sol
+++ b/contracts/Marketplace.sol
@@ -7,11 +7,12 @@ import "@openzeppelin/contracts/utils/structs/EnumerableSet.sol";
 import "./Configuration.sol";
 import "./Requests.sol";
 import "./Proofs.sol";
+import "./SlotReservations.sol";
 import "./StateRetrieval.sol";
 import "./Endian.sol";
 import "./Groth16.sol";
 
-contract Marketplace is Proofs, StateRetrieval, Endian {
+contract Marketplace is SlotReservations, Proofs, StateRetrieval, Endian {
   using EnumerableSet for EnumerableSet.Bytes32Set;
   using Requests for Request;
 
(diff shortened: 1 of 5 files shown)


emizzle:
Closing to rename branch (helps when there will be branches based on branches to prevent a huge PR)

PR #178, opened 2024-09-19
Updates circuit file hash

benbierens:
Circuit repository was changed to create TAR.GZ artifacts instead of ZIP, as part of the downloader-rework.
This means new hash.
This means new deployment.

hardhat and disttest networks were updated.
wasm and r1cs files were updated, but appear to be binary equal.

Do these need updating?
hardhat/example-proofs
hardhat/proof_main.circom
If yes, how do we generate them?


AuHau (approved):
LGTM. As the circuit itself didn't change, those files are not needed to be updated.

PR #185, opened 2024-10-03
feat(slot-reservations): require slots to be reserved before filling slot

### LINKED ISSUE #184: feat(slot-reservations): require slots to be reserved before filling
Revert if slot is not reserved when attempting to fill it.

Check if `_reservations` contains the sender address for the slot.

emizzle:
Closes #184.

This is being implemented as part of a [longer list of tasks for slot reservations](https://github.com/codex-storage/codex-pm/issues/45) to prevent PRs from being too large.

Requires that a slot is reserved before slot is allowed to be filled.

### DIFF
--- a/contracts/Marketplace.sol
+++ b/contracts/Marketplace.sol
@@ -14,6 +14,7 @@ import "./Groth16.sol";
 
 contract Marketplace is SlotReservations, Proofs, StateRetrieval, Endian {
   using EnumerableSet for EnumerableSet.Bytes32Set;
+  using EnumerableSet for EnumerableSet.AddressSet;
   using Requests for Request;
 
   IERC20 private immutable _token;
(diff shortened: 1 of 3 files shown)


markspanbroek (approved):
I expected another test to check that slot cannot be filled if we didn't reserve it first?

emizzle:
Great catch @markspanbroek, thank you. Test added.

0x-r4bbit:
I ran the failing rule in isolation and it seems to pass just fine:

https://prover.certora.com/output/6199/f9e2ebacef7f45758ee1986da6f1ebdd?anonymousKey=c12f7acafcc05fbec15f087c6b2d79aee7446396

Gonna rerun the job on CI to see if it consistently times out. If that is the case, we might need to extract it into its own configuration.

PR #189, opened 2024-10-07
fix(certora): fix failing rule that cancelled requests stay cancelled

0x-r4bbit:
The prover found a counter example where it starts with a request that has an id `0`, which always results in a `SlotState.New`, resulting in that request becoming `RequestState.Started` while it should be `RequestState.Cancelled`.

We're fixing this by requiring that when the `requestId == 0` then the slot's client has to be `0` address as well.

This will cause a reverting case in the prover, preventing it to find a false positive.

### DIFF
--- a/certora/specs/Marketplace.spec
+++ b/certora/specs/Marketplace.spec
@@ -142,6 +142,22 @@ hook Sstore _requestContexts[KEY MarketplaceHarness.RequestId RequestId].endsAt 
 |              Helper functions              |
 --------------------------------------------*/
 
+function ensureValidRequestId(Marketplace.RequestId requestId) {
+    // Without this, the prover will find counter examples with `requestId == 0`,
+    // which are unlikely in practice as `requestId` is a hash from a request object.
+    // However, `requestId == 0` enforces `SlotState.Free` in the `fillSlot` function regardless,
+    // which ultimately results in counter examples where we have a state change
+    // RequestState.Finished -> RequestState.Started, which is forbidden.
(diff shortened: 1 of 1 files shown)


AuHau (approved):

AuHau:
Thanks 🙏

emizzle (commented):
Could we instead change the require to `requestId != 0 && request.client != 0`?

0x-r4bbit:
> Could we instead change the require to requestId != 0 && request.client != 0

@emizzle I remember we've considered this in the past. There was a reason why we decided against it, but I can't remember why.. I'll try again and see if it would still do the trick.

0x-r4bbit:
@emizzle I've updated the PR according to your suggestions and it seems to work fine!

There's three places where we use this rule. Here's individual prover runs for them:

- `allowedRequestStateChanges` - https://prover.certora.com/output/6199/fd25372e06c04fbb8cbf57b0b7b5616d?anonymousKey=f3f799000f1dc0e2d3ca8de71401bd1bd56a0624
- `finishedRequestsStayFinished` - https://prover.certora.com/output/6199/d2a8bca47979470abe79e67a93f17973?anonymousKey=a7cd2976696e930b176ce4e63b4d73121b8eb463
- `cancelledRequestsStayCancelled` - https://prover.certora.com/output/6199/d25a5b14713b445984295a479806a2f0?anonymousKey=ac57b84161fa7c48121a023d85af5e815c0580cd

PR #191, opened 2024-10-08
feat: require proof for freeSlot

### LINKED ISSUE #143: Require proof for `freeSlot`
Given that there is leniency in allowed proofs missed before being slashed, `slashCriterion`, there should be a final proof submitted from SPs at the end of the contract duration before they are paid out?

### Rationale
SPs could delete data for the last `x` periods of the contract, without any penalty at all, where `x` is: `(slashCriterion-1) * proofProbability`. Considering that each period could be 24 hours, there is a potential for SPs to be able to delete their contractually obliged data for quite a few days without penalty.

There are also a few incentive structures in place that rely on the idea that SPs will only get paid out after the contract successfully finishes, so adding a single additional proof requirement at the end would only reinforce the idea that data must be stored until the end to get paid out.

### Implementation
Add an optional parameter, `proof`, to `freeSlot`. On successfully completed requests, SPs must provide a valid proof when calling `freeSlot` before payout is allowed.

emizzle:
Close #143.

To prevent SPs from going unpenalised during the last periods, a proof is now required to be paid out, meaning calls to `freeSlot` for finished and cancelled requests require a storage proof to be provided.

### DIFF
--- a/certora/specs/Marketplace.spec
+++ b/certora/specs/Marketplace.spec
@@ -151,19 +151,19 @@ function canStartRequest(method f) returns bool {
 }
 
 function canFinishRequest(method f) returns bool {
-    return f.selector == sig:freeSlot(Marketplace.SlotId, address, address).selector ||
-        f.selector == sig:freeSlot(Marketplace.SlotId).selector;
+    return f.selector == sig:freeFinishedSlot(Marketplace.SlotId, Marketplace.Groth16Proof, address, address).selector ||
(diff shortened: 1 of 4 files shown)


markspanbroek:
Closing this, because as we discussed elsewhere, this requirement is incompatible with the design of the Vault (#220), whereby a user can always directly withdraw their funds from the vault after the storage request is finished.

PR #193, opened 2024-10-11
feat: repair reward

### LINKED ISSUE #154: Additional reward for host repairing slot
As described in  [the design doc](https://github.com/codex-storage/codex-research/blob/41c4b4409d2092d0a5475aca0f28995034e58d14/design/marketplace.md#repairs), the node that repairs a slot should be awarded additional reward as compensation for its extra work as it needs to download more data in order to reconstruct the original dataset. This reward should come from the collateral of the previous host. Currently, we are not awarding this reward.

Thinking about the size of the repair reward, it is not trivial to decide what it should be. 

The current thinking coming from [the design doc](https://github.com/codex-storage/codex-research/blob/41c4b4409d2092d0a5475aca0f28995034e58d14/design/marketplace.md#repairs) is that it should be a fraction of the collateral, which would be specified on contract constructor, hence same for the whole network. But as that section also states the reward should not be too small nor too big. I would argue that the "fraction of collateral" approach does not satisfy this requirement as the reward could easily become either too small or too big based on the user inputted `collateral` parameter (he might not even realize it affects the repair reward).

If we are targeting to compensate for the required work to reconstruct the lost slot's data, I would argue that the size of the reward  is a function where the size of the slot plays a major role. I would hence suggest another approach which is linked to our general shift for pricing "per byte". If we would have collateral also scoped for "per byte" then the fraction of "per byte" payment would better reflects the final repair reward as it would scale with the size of the dataset. WDYT?

AuHau:
Closes #154 

The reward is attributed using a discount provided for the collateral required to be deposited. In this way, the reward is actually paid out at the end of Storage Request where the full amount of collateral is returned (provided no slashing occurred).

### DIFF
--- a/contracts/Marketplace.sol
+++ b/contracts/Marketplace.sol
@@ -149,28 +149,47 @@ contract Marketplace is SlotReservations, Proofs, StateRetrieval, Endian {
     slot.slotIndex = slotIndex;
     RequestContext storage context = _requestContexts[requestId];
 
-    require(slotState(slotId) == SlotState.Free, "Slot is not free");
+    require(
+      slotState(slotId) == SlotState.Free ||
+        slotState(slotId) == SlotState.Repair,
+      "Slot is not free"
+    );
 
     _startRequiringProofs(slotId, request.ask.proofProbability);
     submitProof(slotId, proof);
 
     slot.host = msg.sender;
-    slot.state = SlotState.Filled;
     slot.filledAt = block.timestamp;
 
     context.slotsFilled += 1;
     context.fundsToReturnToClient -= _slotPayout(requestId, slot.filledAt);
 
     // Collect collateral
-    uint256 collateralAmount = request.ask.collateral;
+    uint256 collateralAmount;
+    if (slotState(slotId) == SlotState.Repair) {
+      // Host is repairing a slot and is entitled for repair reward, so he gets "discounted collateral"
+      // in this way he gets "physically" the reward at the end of the request when the full amount of collateral
+      // is returned to him.
+      collateralAmount =
+        request.ask.collateral -
+        ((request.ask.collateral * _config.collateral.repairRewardPercentage) /
+          100);
+    } else {
+      collateralAmount = request.ask.collateral;
+    }
     _transferFrom(msg.sender, collateralAmount);
     _marketplaceTotals.received += collateralAmount;
-    slot.currentCollateral = collateralAmount;
+    slot.currentCollateral = request.ask.collateral; // Even if he has collateral discounted, he is operating with full collateral
 
     _addToMySlots(slot.host, slotId);
 
+    slot.state = SlotState.Filled;
     emit SlotFilled(requestId, slotIndex);
-    if (context.slotsFilled == request.ask.slots) {
+
+    if (
+      context.slotsFilled == request.ask.slots &&
+      context.state != RequestState.Started // In case of repair, we already have the request started, so we prevent from "restarting it"
+    ) {
       context.state = RequestState.Started;
       context.startedAt = block.timestamp;
       emit RequestFulfilled(requestId);
@@ -298,6 +317,9 @@ contract Marketplace is SlotReservations, Proofs, StateRetrieval, Endian {
     _removeFromMySlots(slot.host, slotId);
     uint256 slotIndex = slot.slotIndex;
     delete _slots[slotId];
+    delete _reservations[slotId]; // We purge all the reservations for the slot
+    slot.state = SlotState.Repair;
+    slot.requestId = requestId;
     context.slotsFilled -= 1;
     emit SlotFreed(requestId, slotIndex);
     _resetMissingProofs(slotId);
--- a/contracts/Marketplace.sol
+++ b/contracts/Marketplace.sol
@@ -188,7 +188,7 @@ contract Marketplace is SlotReservations, Proofs, StateRetrieval, Endian {
 
     if (
       context.slotsFilled == request.ask.slots &&
-      context.state != RequestState.Started // In case of repair, we already have the request started, so we prevent from "restarting it"
+      context.state == RequestState.New // Only New requests can "start" the requests
     ) {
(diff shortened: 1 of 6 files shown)


contracts/Marketplace.sol
```diff
@@ -149,28 +149,47 @@ contract Marketplace is SlotReservations, Proofs, StateRetrieval, Endian {
...
-    uint256 collateralAmount = request.ask.collateral;
+    uint256 collateralAmount;
+    if (slotState(slotId) == SlotState.Repair) {
+      // Host is repairing a slot and is entitled for repair reward, so he gets "discounted collateral"
+      // in this way he gets "physically" the reward at the end of the request when the full amount of collateral
+      // is returned to him.
+      collateralAmount =
+        request.ask.collateral -
+        ((request.ask.collateral * _config.collateral.repairRewardPercentage) /
+          100);
```
emizzle, line 176: 🟡  IIRC, this is used in several places and would be nice to create a function, eg `repairReward()`

contracts/Marketplace.sol
```diff
@@ -149,28 +149,47 @@ contract Marketplace is SlotReservations, Proofs, StateRetrieval, Endian {
...
     _addToMySlots(slot.host, slotId);
 
+    slot.state = SlotState.Filled;
     emit SlotFilled(requestId, slotIndex);
-    if (context.slotsFilled == request.ask.slots) {
+
+    if (
+      context.slotsFilled == request.ask.slots &&
+      context.state != RequestState.Started // In case of repair, we already have the request started, so we prevent from "restarting it"
+    ) {
```
emizzle, line 192: What is this event being used for in the case of repair? 

Emitting this event in the case of repair will cause Codex to re-add all the slots in the request. Currently, Codex already handles the `SlotFreed` event, re-adding that slot to the queue.

contracts/Marketplace.sol
```diff
@@ -298,6 +317,9 @@ contract Marketplace is SlotReservations, Proofs, StateRetrieval, Endian {
     _removeFromMySlots(slot.host, slotId);
     uint256 slotIndex = slot.slotIndex;
     delete _slots[slotId];
+    delete _reservations[slotId]; // We purge all the reservations for the slot
+    slot.state = SlotState.Repair;
+    slot.requestId = requestId;
```
emizzle, line 322: Why is this needed? Shouldn't it have been already set in `fillSlot`?

emizzle (commented):
Nice job, one comment regarding the re-emission of `RequestFulfilled`...

contracts/Marketplace.sol
AuHau: In contracts it is used only here. There is the test counterpart, but I would keep it explicitly "reimplemented" there.

contracts/Marketplace.sol
AuHau: Not sure if I follow your question.

This change **prevents** from emitting `RequestFulfilled` when repaired the "last slot" (eq. when only one slot failed and was repaired in which case we would get the event, but I am preventing that here).

There is no event emitted when slot is repaired.

contracts/Marketplace.sol
AuHau: `delete _slots[slotId]` will delete all the info ;-) so I am just repopulating required data here.

We need the `requestId` otherwise the `slotState()` will report `Free` even though `slot.state` is set.

contracts/Marketplace.sol
emizzle: Ok, when I first reviewed this, I think I thought that the RequestState would change as well when repairing, which obviously does not make sense.

Would it be better to further restrict this to `context.state == RequestState.New`? I'm thinking that if a request has already ended (`RequestState.Finished` or even `Failed`) with one or more of the slots in the `Repair` state, then someone could potentially call `fillSlot` and emit this event (it would, of course incur, a cost, but the "attack" could be disruption of the network).

emizzle (approved):
Nice work! I had one comment (https://github.com/codex-storage/codex-contracts-eth/pull/193/files#r1819895466), but will approve now as it's not a blocker.

AuHau:
I realized there are two paths here where I took one way "automatically," but it might be worth considering the other option as well.

1) **collateral discount** - because I originally had in mind that when SP repairs a slot, it will get the repair reward at the point of filling that slot, and providing proof for it. Since he also needs to deposit collateral, I just thought that these two flows could "nullify" into the form of a discount on collateral. Then at the end of the successfully finished Request, the SP would get the whole collateral sum back and hence also "the repair reward".

This has a few consideration - if the Request fails the reward is never paid, although the SP had the discount on collateral, so "he loses less". Also, this weakens the guarantees of collateral role as repairing SPs are not so vested.

2) **repair reward** - An alternative could be that the SP would have to deposit the full collateral and the repair reward would be added at the end of (successfully finished?) Request.

This would not weaken the collateral guarantees, but it might lessen the motivation of SPs to repair? Because of the market's uncertainty, that request will be successfully finished. It might even happen that the repairing SPs will be also kicked out from the slot because of whatever reasons and then the repair reward would be voided.

markspanbroek (approved):
Nice, thanks @AuHau!

emizzle:
> This has a few consideration - if the Request fails the reward is never paid, although the SP had the discount on collateral, so "he loses less". Also, this weakens the guarantees of collateral role as repairing SPs are not so vested.

It's good that you were thinking about this! Yes, I agree that this could weaken the guarantees, especially if there are a lot of repairs in a single storage request. 

> it would make sense to payout the reward "no matter what" after the slot is repaired and re-filled.

Having SPs getting paid the repair reward at the end of a successful storage request would incentivise them further to keep the storage request running.

>  It might even happen that the repairing SPs will be also kicked out from the slot because of whatever reasons and then the repair reward would be voided.

Possibly, but they also have a chance to repair slots and keep the contract running.

Overall, my vote would be to use a *repair reward* and pay out the SPs at the end of the (successful) contract.

PR #196, opened 2024-10-17
fix(slot-reservations): ensure slot is free

emizzle:
Ensure that the slot state is free before allowing reservations


AuHau (approved):
Good catch.

PR #198, opened 2024-10-28
renames config to configuration

emizzle:
This is not the right reason to be making this kind of change, but a very hard to debug symbol clash in codex for `config`. Changing this to `configuration` is the easiest way to fix the issue.

I spent nearly an entire day unsuccessfully trying to solve a strange symbol clash with `config` inside the integration tests. This was a last resort, and it works. Alternative attempts to solve the clash are welcome.

This is part of the getting the latest ethers merged into codex (to fix the nonce issue plaguing the testnet).

### DIFF
--- a/contracts/Marketplace.sol
+++ b/contracts/Marketplace.sol
@@ -88,7 +88,7 @@ contract Marketplace is SlotReservations, Proofs, StateRetrieval, Endian {
     _config = configuration;
   }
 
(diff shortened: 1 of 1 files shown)


AuHau (approved):
wow, yeah I can imagine how mind-fu**** bug this could have been 😳 I don't see any problem with renaming it 👍

PR #202, opened 2024-11-14
Increase proof period to 90 seconds on hardhat network

markspanbroek:
Reason: on windows, the nim-codex integration tests run so slow that they cannot fill a slot within 60 seconds.
Slowness is most likely caused by https://github.com/codex-storage/nim-ethers/issues/95.


markspanbroek:
Sorry, merging this one without review :smile:

PR #205, opened 2024-12-04
ci: tweaks run triggers and checkout hash

AuHau:
Tweaks the CI trigger limiting "push" trigger only to `master` and also configures to checkout the true PR branch.


veaceslavdoina (approved):
Thanks, we discussed that already :)

PR #208, opened 2025-01-10
Feat: price per byte

marcinczenko:
> This PR is one of the PRs for issue https://github.com/codex-storage/nim-codex/issues/460. See https://github.com/codex-storage/nim-codex/issues/460#issuecomment-2583250768.

Here we do the following:

1. changes `ask.reward` to `ask.pricePerByte`
2. changes `ask.collateral` to `ask.collateralPerByte`
3. updates all related computations and tests

### DIFF
--- a/contracts/Marketplace.sol
+++ b/contracts/Marketplace.sol
@@ -127,6 +127,10 @@ contract Marketplace is SlotReservations, Proofs, StateRetrieval, Endian {
     return _token;
   }
 
+  function currentCollateral(SlotId slotId) public view returns (uint256) {
+    return _slots[slotId].currentCollateral;
+  }
+
   function requestStorage(Request calldata request) public {
     RequestId id = request.id();
 
--- a/contracts/Marketplace.sol
+++ b/contracts/Marketplace.sol
@@ -586,7 +586,10 @@ contract Marketplace is SlotReservations, Proofs, StateRetrieval, Endian {
     Request storage request = _requests[requestId];
     if (startingTimestamp >= endingTimestamp)
       revert Marketplace_StartNotBeforeExpiry();
-    return (endingTimestamp - startingTimestamp) * request.ask.reward;
+    return
+      (endingTimestamp - startingTimestamp) *
+      request.ask.pricePerByte *
+      request.ask.slotSize;
   }
 
   function getHost(SlotId slotId) public view returns (address) {
--- a/contracts/Marketplace.sol
+++ b/contracts/Marketplace.sol
@@ -76,7 +76,7 @@ contract Marketplace is SlotReservations, Proofs, StateRetrieval, Endian {
     uint256 filledAt;
     uint256 slotIndex;
     /// @notice Tracks the current amount of host's collateral that is to be payed out at the end of Slot's lifespan.
-    /// @dev When Slot is filled, the collateral is collected in amount of request.ask.collateral
+    /// @dev When Slot is filled, the collateral is collected in amount of request.ask.collateralPerByte * request.ask.slotSize
     /// @dev When Host is slashed for missing a proof the slashed amount is reflected in this variable
     uint256 currentCollateral;
     address host; // address used for collateral interactions and identifying hosts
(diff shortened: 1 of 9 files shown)


contracts/Marketplace.sol
```diff
@@ -284,7 +284,8 @@ contract Marketplace is SlotReservations, Proofs, StateRetrieval, Endian {
     // TODO: Reward for validator that calls this function
 
     if (missingProofs(slotId) % _config.collateral.slashCriterion == 0) {
-      uint256 slashedAmount = (request.ask.collateral *
+      uint256 slashedAmount = (request.ask.collateralPerByte *
+        request.ask.slotSize *
```
emizzle, line 288: 🟡  Very much just a suggestion here. Since we use `request.ask.collateralPerByte * request.ask.slotSize` in a few places, we could introduce a calculated property, `collateral`, so that we could use `request.ask.collateral` instead. This can be with a [`using..for`](https://docs.soliditylang.org/en/latest/contracts.html#using-for):

```solidity
// Requests.sol
using {collateral} for Ask global;

function collateral(Ask ask) public view returns (uint256) {
  return ask.collateralPerByte * ask.slotSize;
}

// Marketplace.sol
uint256 slashedAmount = (request.ask.collateral * _config.collateral.slashPercentage) / 100;
```

Would be similar for pricePerByte. Just a suggestion, not sure how well it would work or if it's too much extra work.

emizzle (approved):
Nice job 👍

contracts/Marketplace.sol
marcinczenko: Please take a look at the last commit - I have added `AskHelpers` library inside of the `Requests.sol` next to the `Requests` library. Both `Requests` library and `Marketplace.sol` use `AskHelpers` to compute total `price()` and `collateral()` for the slot.

AuHau (commented):
Few suggestions from my side ;-)

contracts/Requests.sol
```diff
@@ -45,7 +45,19 @@ enum SlotState {
   Repair // when slot slot was forcible freed (host was kicked out from hosting the slot because of too many missed proofs) and needs to be repaired
 }
 
+library AskHelpers {
+  function collateral(Ask memory ask) internal pure returns (uint256) {
+    return ask.collateralPerByte * ask.slotSize;
+  }
+
+  function price(Ask memory ask) internal pure returns (uint256) {
```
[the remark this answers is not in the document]
marcinczenko: I will make names more descriptive. I've listed all the current names in https://github.com/codex-storage/nim-codex/issues/460#issuecomment-2598209735.

deploy/token.js
```diff
@@ -1,16 +1,31 @@
-const MINTED_TOKENS = 1_000_000_000
+const MINTED_TOKENS = 1_000_000_000_000_000
```
emizzle, line 1: I just noticed this change. Was it meant to be changed? 1 quadrillion tokens?

emizzle (commented):
I think there may have been a leftover change in the number of tokens deployed?

contracts/Marketplace.sol
emizzle: I don't have strong opinions either way, but I'm curious why you chose to use a new library `AskHelpers` instead of including the functions in the `Request` library?

deploy/token.js
marcinczenko: Yes, the amounts that need to be transferred in `nim-codex`, `testContracts` are close to that in some cases and we using the network from `vendor/codex-contracts-eth` when running those tests. We could probably do it also pragmatically, but this way seemed to be the most convenient. This is also by the way, the reason, I started to have reservations in using higher random values for `pricePerBytePerSecond` and `collateralPerByte`.

contracts/Marketplace.sol
marcinczenko: I somehow felt that the `Requests` library is meant to extend the type `Request`, in which case you can call e.g. `request.id()` and have `function id(Request memory request) internal pure returns (RequestId)` from the `Requests` library called. We do seem to have functions in the `Request` library that do not extend the `Request` type, e.g. `slotId` which takes `RequestId` as first argument, so that we use it like `Requests.slotId`. I somehow like when library is dedicated to one type and so I did not want to further "pollute" the `Requests` library with `Ask` helpers. So I do not have strong opinion either, but somehow felt this will be more elegant.

deploy/token.js
marcinczenko: I actually, got concerned about those values myself, so I had to quickly check how does all this fit ERC20 and our use of tokens, and what I computed is perfectly fine, especially that we will also perhaps have like 18 decimals.

marcinczenko: And I also thought to mint programmatically on per-test basis, but since we had already everything in place and we used it like this before I preferred to just increase the amount of tokens minted...

PR #209, opened 2025-01-14
Reward validator when marking missing proof

### LINKED ISSUE #70: Reward for validators that correctly mark proof as missing
Currently in the `markProofAsMissing` there is no reward for the node that calls this function correctly. 

It is a question when and whom should get paid the reward as not every call to `markProofAsMissing` leads to slashing (from which the funds for reward could be deducted) and in the same time if only those calls that leads to slashing will yield the reward, then there is no incentive to really call this function for the other cases...

2-towns:
Fix #70


emizzle (approved):
LGTM, well done!

PR #210, opened 2025-01-14
Remove missing proof leniency

### LINKED ISSUE #195: Remove missing proof leniency
Currently, SPs are not slashed if they miss up to `slashCriterion` proofs. Once `slashCriterion` proofs are missed, the SP is slashed. The problem this creates is that there is no source of funds for validators that marked the proofs as missing when the SP was not slashed.

To simplify this situation, remove the leniency of allowing `slashCriterion` proofs to be missed before slashing occurs.

2-towns:
Fix #195

### DIFF
--- a/configuration/configuration.js
+++ b/configuration/configuration.js
@@ -6,7 +6,6 @@ const DEFAULT_CONFIGURATION = {
   collateral: {
     repairRewardPercentage: 10,
     maxNumberOfSlashes: 2,
-    slashCriterion: 2,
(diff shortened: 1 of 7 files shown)


emizzle (approved):
LGTM, well done Arnaud 👏

AuHau:
@dryajov I am not sure if we checked with you if this approach (eq. removing some of the "laziness" in the system) is OK with you. WDYT?

PR #211, opened 2025-01-15
Assert than secondsPerPeriod cannot be zero

2-towns:


emizzle (approved):
LGTM

AuHau:
Same here ;-) 

https://github.com/codex-storage/nim-codex/blob/master/codex/contracts/marketplace.nim

PR #212, opened 2025-01-15
Reject when probability is 0

### LINKED ISSUE #105: Zero probability should be rejected
https://github.com/codex-storage/codex-contracts-eth/blob/57e8cd5013325f05e16833a5320b575d32a403f3/contracts/Proofs.sol#L40

2-towns:
Fix #105


emizzle (approved):
LGTM

AuHau:
@2-towns please don't forget to update the `nim-codex` with adding the new error type. Otherwise `nim-codex` will fail with non-descriptive error as it won't be able to decode the revert error. Look here: https://github.com/codex-storage/nim-codex/blob/master/codex/contracts/marketplace.nim

PR #215, opened 2025-01-17
Remove the mapping _probabilities

### LINKED ISSUE #137: Remove unneeded `_probabilities`?
When a slot is filled, the proof probability is stored in a private mapping `_probabilities`. However, this is likely not needed as the proof probability for a `Request` is already stored in `Request.Ask`.

This may require passing in the request Id.

2-towns:
Fix #137

### DIFF
--- a/test/Marketplace.test.js
+++ b/test/Marketplace.test.js
@@ -58,18 +58,24 @@ describe("Marketplace constructor", function () {
     await revert()
   })
 
-  function testPercentageOverflow(property) {
+  function testPercentageOverflow(property, expectedError) {
     it(`should reject for ${property} overflowing percentage values`, async () => {
       config.collateral[property] = 101
 
       await expect(
         Marketplace.deploy(config, token.address, verifier.address)
-      ).to.be.revertedWith("Must be less than 100")
+      ).to.be.revertedWith(expectedError)
     })
   }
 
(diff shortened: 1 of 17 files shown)


test/Marketplace.test.js
```diff
@@ -1136,6 +1136,28 @@ describe("Marketplace", function () {
...
+    beforeEach(async function () {
+      switchAccount(client)
+      await token.approve(marketplace.address, maxPrice(request))
+      await marketplace.requestStorage(request)
+      switchAccount(host)
+      await token.approve(marketplace.address, request.ask.collateral)
+    })
+
+    it("calculates correctly the slot probability", async function () {
+      await waitUntilStarted(marketplace, request, proof, token)
```
emizzle, line 1149: Out of curiosity, is this required for the test? I think maybe the slot probability is calculable even if we're not requiring proofs (in fillSlot).

emizzle (approved):
Nice work! I left a small comment/suggestion, but nothing blocking, so will approve 👍

test/Marketplace.test.js
2-towns: Yes you are right, the fill slot is only required for the calculation, we don't need to require proofs.

2-towns: I pushed this change.

PR #217, opened 2025-01-23
test(certora): payout harness

AuHau:
This is PR that adds calculation utilities for the `MarketplaceHarness` in order to write a rule that will validate the sufficiency of Marketplace's balance to payout all required funds.

### DIFF
--- a/certora/harness/MarketplaceHarness.sol
+++ b/certora/harness/MarketplaceHarness.sol
@@ -6,23 +6,145 @@ import {IERC20} from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 import {IGroth16Verifier} from "../../contracts/Groth16.sol";
 import {MarketplaceConfig} from "../../contracts/Configuration.sol";
 import {Marketplace} from "../../contracts/Marketplace.sol";
-import {RequestId, SlotId} from "../../contracts/Requests.sol";
-import {Requests} from "../../contracts/Requests.sol";
+import {Requests, Request, RequestState, SlotState, RequestId, SlotId} from "../../contracts/Requests.sol";
 
 contract MarketplaceHarness is Marketplace {
-    constructor(MarketplaceConfig memory config, IERC20 token, IGroth16Verifier verifier)
-        Marketplace(config, token, verifier)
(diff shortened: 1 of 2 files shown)


certora/harness/MarketplaceHarness.sol
```diff
@@ -6,23 +6,145 @@ import {IERC20} from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
...
+    MarketplaceConfig memory config,
+    IERC20 token,
+    IGroth16Verifier verifier
+  ) Marketplace(config, token, verifier) {}
+
+  function publicPeriodEnd(Period period) public view returns (uint256) {
+    return _periodEnd(period);
+  }
+
+  function slots(SlotId slotId) public view returns (Slot memory) {
```
emizzle, line 24: Should this be called `slot` instead of `slots`?

emizzle (approved):
I know we said we will see if this is sufficient for vac/certora to use in their formal verification, but there's no changes in here that I see as blocking. Also, we should probably wait on @markspanbroek 's Vault changes as he said he introduced the concept of balances.

certora/harness/MarketplaceHarness.sol
AuHau: Well not really part of this PR, but yeah 😅

0x-r4bbit (approved):

certora/harness/MarketplaceHarness.sol
```diff
@@ -6,23 +6,145 @@ import {IERC20} from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
...
+  ) public view returns (uint256) {
+    Request memory request = _requests[requestId];
+    RequestState state = requestState(requestId);
+
+    if (
+      state == RequestState.New ||
+      state == RequestState.Started ||
+      state == RequestState.Failed
+    ) {
+      return request.maxPrice();
```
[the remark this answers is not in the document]
AuHau: Currently yes, if Request is Failed (eq. too much slots failed), then client is refunded the whole amount. This might change soon thanks to the Vault work.

PR #218, opened 2025-02-06
simplify time-based logic in tests, and fix requestEnd()

markspanbroek:
- use the `allowBlocksWithSameTimestamp` hardhat option
- remove block time gymnastics from marketplace tests
- fix erroneous implementation of requestEnd() which surfaced because of the the improved tests


emizzle (approved):
Much more clear this way, thanks Mark. Also, great to finally get the requestEnd calculation fixed after all this time. It wasn't even on my radar that we forgot to update it. 👍

AuHau (approved):
Great improvement! Thanks!

PR #226, opened 2025-03-13
Status L2 Linea test deployment

2-towns:


emizzle:
Hmm, maybe we should commit these files so we can re-deploy on Linea if needed? IOW, maybe we should take this out of draft? Thoughts?

emizzle (commented):
Overall I think this looks good. However, is there a particular reason for wanting to commit the zkey and r1cs? These are both very large files (and could get a lot larger).

2-towns:
> Overall I think this looks good. However, is there a particular reason for wanting to commit the zkey and r1cs? These are both very large files (and could get a lot larger).

There is no specific reason, I wasn't sure the importance to add them into git or not.

2-towns:
> Overall I think this looks good. However, is there a particular reason for wanting to commit the zkey and r1cs? These are both very large files (and could get a lot larger).

Those files are committed for other networks, like testnet for example.

emizzle:
> Those files are committed for other networks, like testnet for example.

Ah yes, I remember we moved them from `nim-codex` to `codex-contracts-eth`. But after the circuit downloader was put in, do we need to keep these files in `codex-contracts-eth`? @dryajov ?

veaceslavdoina:
> > Those files are committed for other networks, like testnet for example.
> 
> Ah yes, I remember we moved them from `nim-codex` to `codex-contracts-eth`. But after the circuit downloader was put in, do we need to keep these files in `codex-contracts-eth`? @dryajov ?

@emizzle, I would assume these files (zkey and r1cs) are similar to the existing network (`codex_testnet` / `codex_devnet`) from my PR #135.

I was added them following an example for [`codexdisttestnetwork`](https://github.com/codex-storage/codex-contracts-eth/commits/master/verifier/networks/codexdisttestnetwork) added by @AuHau in #89, @benbierens in #96 / #178,  and @gmega in #121.

@AuHau was first - let's ask him for more details?

PR #227, opened 2025-03-13
fix: only slots host should be able to submit proof

### LINKED ISSUE #151: Proof submission from other address?
While writing the spec, I realized that we do not check the address of the tx sender for submitting proof. 
Should we restrict that? While I don't necessarily see any problems with leaving it open, my intuition says we should restrict it. Restricting also does not affect anything IMHO, but may prevent some potential issues?

AuHau:
Closes #151


emizzle (approved):
Nice work.

I believe we originally didn't include this on purpose because if someone wants to submit a proof for another host, it was highly disincentivised situation.

PR #229, opened 2025-04-03
chore: add a new canMarkProofAsMissing function

2-towns:
This PR fixes https://github.com/codex-storage/nim-codex/issues/1153 by adding a new function in the contract, `canProofBeMarkedAsMissing`.

Since this function is a view and does not transfer reward funds to the validator, we should no longer encounter the `ERC20: transfer to the zero address` error.

Moreover, it updates the contracts since the `requestEnd` has been fixed in the contracts.

### DIFF
--- a/contracts/Marketplace.sol
+++ b/contracts/Marketplace.sol
@@ -335,7 +335,7 @@ contract Marketplace is SlotReservations, Proofs, StateRetrieval, Endian {
     _proofReceived(id, proof, pubSignals);
   }
 
-  function canProofBeMarkedAsMissing(
+  function canMarkProofAsMissing(
     SlotId slotId,
     Period period
   ) public view slotAcceptsProofs(slotId) {
--- a/contracts/Marketplace.sol
+++ b/contracts/Marketplace.sol
@@ -335,11 +335,19 @@ contract Marketplace is SlotReservations, Proofs, StateRetrieval, Endian {
     _proofReceived(id, proof, pubSignals);
   }
 
-  function markProofAsMissing(SlotId slotId, Period period) public {
-    if (slotState(slotId) != SlotState.Filled)
-      revert Marketplace_SlotNotAcceptingProofs();
+  function canProofBeMarkedAsMissing(
+    SlotId slotId,
+    Period period
+  ) public view slotIsFilled(slotId) {
+    _canMarkProofAsMissing(slotId, period);
+  }
 
+  function markProofAsMissing(
+    SlotId slotId,
+    Period period
+  ) public slotIsFilled(slotId) {
     _markProofAsMissing(slotId, period);
+
     Slot storage slot = _slots[slotId];
     Request storage request = _requests[slot.requestId];
 
@@ -349,7 +357,10 @@ contract Marketplace is SlotReservations, Proofs, StateRetrieval, Endian {
     uint256 validatorRewardAmount = (slashedAmount *
       _config.collateral.validatorRewardPercentage) / 100;
     _marketplaceTotals.sent += validatorRewardAmount;
-    assert(_token.transfer(msg.sender, validatorRewardAmount));
+
+    if (!_token.transfer(msg.sender, validatorRewardAmount)) {
+      revert Marketplace_TransferFailed();
+    }
 
     slot.currentCollateral -= slashedAmount;
     if (missingProofs(slotId) >= _config.collateral.maxNumberOfSlashes) {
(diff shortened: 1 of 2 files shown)


contracts/Marketplace.sol
```diff
@@ -335,11 +335,19 @@ contract Marketplace is SlotReservations, Proofs, StateRetrieval, Endian {
     _proofReceived(id, proof, pubSignals);
   }
 
-  function markProofAsMissing(SlotId slotId, Period period) public {
-    if (slotState(slotId) != SlotState.Filled)
-      revert Marketplace_SlotNotAcceptingProofs();
+  function canProofBeMarkedAsMissing(
```
emizzle, line 338: Just for consistency, maybe we could also call this `canMarkProofAsMissing` (without the leading _ since it's public)?

2-towns: Sorry I didn't get your point. This method, in Marketplace.sol, is public and is named `canProofBeMarkedAsMissing`. Did you confuse with `_canProofBeMarkedAsMissing` in Proofs.sol ?

contracts/Marketplace.sol
```diff
@@ -335,11 +335,19 @@ contract Marketplace is SlotReservations, Proofs, StateRetrieval, Endian {
...
+    SlotId slotId,
+    Period period
+  ) public view slotAcceptsProofs(slotId) {
```
markspanbroek, line 341: I'm not a big fan of view functions that revert. The revert reason is often hard to get to, and it just feels weird that a view on the blockchain state would revert, because there's actually no transaction to revert.

I understand why you did this here, because we already had the logic in `_markProofAsMissing` that reverts when a proof cannot be marked as missing. But I'd rather see a boolean return value on this view function, and no reverts. I'll include a suggestion below as to how this would look in `_markProofAsMissing`.

contracts/Proofs.sol
```diff
@@ -224,16 +224,23 @@ abstract contract Proofs is Periods {
...
+    SlotId id,
+    Period missedPeriod
+  ) internal view {
     uint256 end = _periodEnd(missedPeriod);
     if (end >= block.timestamp) revert Proofs_PeriodNotEnded();
     if (block.timestamp >= end + _config.timeout)
       revert Proofs_ValidationTimedOut();
     if (_received[id][missedPeriod]) revert Proofs_ProofNotMissing();
     if (!_isProofRequired(id, missedPeriod)) revert Proofs_ProofNotRequired();
     if (_missing[id][missedPeriod]) revert Proofs_ProofAlreadyMarkedMissing();
```
markspanbroek, line 243: Here's how you can have a boolean return value. The disadvantage is that you don't get a detailed reason why the proof could not be marked as missing, but I think that that's ok.

```suggestion
    require(_canMarkProofAsMissing(id, missedPeriod), Proofs_ProofNotMissing());

    _missing[id][missedPeriod] = true;
    _missed[id] += 1;
  }

  function _canMarkProofAsMissing(
    SlotId id,
    Period missedPeriod
  ) internal view returns (bool) {
    uint256 end = _periodEnd(missedPeriod);
    if (end >= block.timestamp) return false;
    if (block.timestamp >= end + _config.timeout)
      return false;
    if (_received[id][missedPeriod]) return false;
    if (!_isProofRequired(id, missedPeriod)) return false;
    if (_missing[id][missedPeriod]) return false;
    return true;
```

markspanbroek (commented):
Thanks @2-towns, I left some comments.

contracts/Proofs.sol
2-towns: I feel that having the reason why a proof cannot be marked as missing is important, especially for the testnet, but maybe I am wrong.

emizzle: Do we know if Linea will report custom errors on the testnet?

2-towns: I am pretty sure it does but I have asked cyp for confirmation

emizzle: If it does, I'd be in favour of reverting with more information. I do completely agree that it feels weird to revert a `view` function, but I prefer to live to with that weirdness if it makes it easier to debug why a proof was not able to be marked as missing. That part of the logic is so complex, that any assistance debugging pays off in dividends. I've personally spent hours trying to debug this the hard way with integration test logs and it's not very fun :)

emizzle (approved):

contracts/Proofs.sol
2-towns: Okay so I got the confirmation from cyp, it is supported. Here is a transaction example: https://lineascan.build/tx/0x9a32b515d5e1f86f941ead119da6f220391fc6fa56e6d4a23a30b537ed7f595a

emizzle:
Good to go (with a rebase) as long as @markspanbroek is ok with the reverting view function?

PR #231, opened 2025-04-09
feat: move to hardhat ignition for deployments

2-towns:
The principal goal of this PR is to move to Hardhat Ignition for deployments, but it includes several changes:

1- Update to ethers 6 
2- Replace `hardhat-waffle` by `@nomicfoundation/hardhat-toolbox` 
3- Run tests in parallel
4- Add a npm command for coverage
5- Add a npm command for gas report 
6- Update to Node 22 on CI 
7- Update dependencies to the last versions except chai because version 4 is required by hardhat
8- Fix security errors using `npm audit fix`


.github/workflows/ci.yml
```diff
@@ -29,7 +28,7 @@ jobs:
       # workaround for https://github.com/NomicFoundation/hardhat/issues/3877
```
veaceslavdoina, line 28: We could remove that comment probably as it was related to the Node 18.15.

2-towns: Oh yes, thanks !

veaceslavdoina:
And one earlier PR - https://github.com/codex-storage/codex-contracts-eth/pull/65, we probably cloud close it?

package.json
```diff
@@ -2,30 +2,29 @@
   "name": "codex-contracts-eth",
   "license": "MIT",
   "scripts": {
-    "test": "npm run lint && hardhat test",
+    "test": "npm run lint && hardhat test --parallel",
     "fuzz": "hardhat compile && fuzzing/fuzz.sh",
-    "start": "hardhat node --export deployment-localhost.json",
+    "start": "hardhat node",
```
emizzle, line 7:
```suggestion
    "start": "hardhat node && npm run deploy",
```

Previously, doing `rm -rf deployments/localhost && npm start`, I would see:
```shell
> hardhat node --export deployment-localhost.json

Nothing to compile
Minting 1000000000000000 tokens to address 0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266
...

Deployed Marketplace with Groth16 Verifier at:
0x322813Fd9A801c5507c9de605d63CEA4f2CE6c44

Deployed Marketplace with Test Verifier at:
0xa85233C63b9Ee964Add6F2cffe00Fd84eb32338f

Started HTTP and WebSocket JSON-RPC server at http://127.0.0.1:8545/

Accounts
========

WARNING: These accounts, and their private keys, are publicly known.
Any funds sent to them on Mainnet or any other live network WILL BE LOST.

Account #0: 0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266 (10000 ETH)
Private Key: 0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80
...
```

But with this branch, if I remove `deployments/localhost`, it doesn't seem to get recreated with `npm start`. Also, I don't see "Deployed Marketplace with Groth16 Verifier at" in the console. Is this expected? My guess is the contracts deployed are not actually deployed with `npm start`, eg running `make testContracts` in codex while running a node from this branch:
```shell
setup] Unhandled error: unable to decode return value as (MarketplaceConfig,) [ContractError]
```
and in the hardhat output I see:
```shell
eth_call
  WARNING: Calling an account which is not a contract
  From: 0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266
  To:   0xa85233c63b9ee964add6f2cffe00fd84eb32338f
```

So if we run the node, then deploy the contracts, I think that would be more in line with expected behaviour.

emizzle: I tested `hardhat node && npm run deploy` but it didn't have the expected output, ie I didn't see
```shell
Deployed Marketplace with Groth16 Verifier at:
0x322813Fd9A801c5507c9de605d63CEA4f2CE6c44

Deployed Marketplace with Test Verifier at:
0xa85233C63b9Ee964Add6F2cffe00Fd84eb32338f
```
I tried launching the node with `npm start`, then running `npm run deploy` in a separate term, but that also didn't work.

So I'm not entirely sure if contracts are not being deployed.

emizzle (commented):
Looking good, Arnaud. A couple of questions:

It seems that the scripts in `ignition/modules` have replaced the deploy scripts in `deploy`? Should we remove the `deploy` dir with `marketplace.js`, `token.js`, and `verifier.js`?

Hardhat documentation mentions that the [deployment artefacts should be created in `ignition/deployments`](https://hardhat.org/ignition/docs/guides/deploy#deployment-artifact-folders):
1. Should we remove the `deployments` directory since it's old?
2. This directory is not being created for me when I run `npm start` and/or `npm run deploy`. My initial guess is that this is due to deployment via a script, and couldn't find any immediate suggestion on how to rectify.

emizzle:
> But when you run hardhat ignition deploy ignition/modules/marketplace.js you have to pass --network localhost if you want to use the hardhat node started. Otherwise, it will use an in memory instance.

Just wanted to say, unless something has changed in v6, I believe the hardhat network is always an in-memory network.

2-towns:
> Just wanted to say, unless something has changed in v6, I believe the hardhat network is always an in-memory network.

Yes, when you run `npx hardhat node`, it will start an in-memory network. 

But for hardhat ignition, if you run `hardhat ignition deploy ignition/modules/marketplace.js`, it will run a in-memory deployment, like a dry-run, so the changes will not be reflected on the hardhat in-memory network. You need to run  `hardhat ignition deploy ignition/modules/marketplace.js --network localhost` in order to `really` deploy on the hardhat in-memory network.

package.json
2-towns: Okay so I slightly changed the start command to ` "concurrently \"hardhat node\" \"sleep 2 &&  npm run deploy\""`. Now the contracts should be deployed and you should see the `deployments` folder in ignition folder. 

But...you will need be able to test in nim-codex (e.g, `make testContracts`) because [increaseAllowance was removed from openzeppelin](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4585). I will create new PRs for that.

emizzle:
Was thinking further about deleting the old deployments folder. They contain information about previous deployments to networks like Taiko and Testnet. Should we try to migrate them to `ignition/deployments` somehow instead of deleting them?

2-towns:
> Was thinking further about deleting the old deployments folder. They contain information about previous deployments to networks like Taiko and Testnet. Should we try to migrate them to `ignition/deployments` somehow instead of deleting them?

What we can do is having a small script like this: 

```js
module.exports = buildModule("Token", (m) => {
  const previousJsonFile = path.join(__dirname, "./TestToken.json")
  const artifact = JSON.parse(fs.readFileSync(previousJsonFile, "utf8"))
  const address = artifact.address
  const token = m.contractAt("TestToken", address, {})
  return { token }
})
```

Then you can run 

```bash
npx hardhat ignition deploy ignition/modules/migration/token.js  --network taiko_test
```

It will recreate the deployment files for the contract address. The contract ABI may differ because OpenZeppelin has been updated, so we may want to keep the previous contract ABI by updating the generated artifacts. Additionally, we could keep the metadata for historical reference, but it is no longer used by ignition.

package.json
emizzle: Can you explain why `"concurrently \"hardhat node\" \"sleep 2 &&  npm run deploy\""` works but `hardhat node && npm run deploy` does not? It seems odd that they'd work "concurrently" but not separately?

Also, I noticed you added a `predeploy` npm script. When does this get executed? I couldn't find anything in the documentation about this. Perhaps if there's a `predeploy`, then there'd be a `postdeploy` that would fit our needs here?

ignition/modules/token.js
```diff
@@ -0,0 +1,24 @@
...
+
+  if (hre.network.config && hre.network.config.tags.includes("local")) {
+    for (let i = 0; i < MAX_ACCOUNTS; i++) {
+      const account = m.getAccount(i)
+      m.call(token, "mint", [account, MINTED_TOKENS], {
+        from: deployer,
+        id: `SendingEth_${i}`,
+      })
+    }
+  }
```
emizzle, line 21: I noticed that each of the 20 accounts has 10000 ETH in there without running this script. Perhaps we no longer need it?

package.json
```diff
@@ -2,31 +2,32 @@
...
+    "start": "concurrently \"hardhat node\" \"sleep 2 &&  npm run deploy\"",
     "compile": "hardhat compile",
-    "format": "prettier --write contracts/**/*.sol test/**/*.js",
-    "format:check": "prettier --check contracts/**/*.sol test/**/*.js",
+    "format": "prettier --write test/**/*.js --plugin=prettier-plugin-solidity contracts/**/*.sol ",
+    "format:check": "prettier --check test/**/*.js --plugin=prettier-plugin-solidity contracts/**/*.sol",
     "lint": "solhint contracts/**.sol",
-    "deploy": "hardhat deploy",
+    "deploy": "hardhat ignition deploy ignition/modules/marketplace.js --network localhost",
+    "predeploy": "hardhat run scripts/mine.js --network localhost",
```
emizzle, line 13: When I tried running hardhat against the network `localhost`, I got an error that this was not the right network:
```bash
❯ npx hardhat node --network localhost
Error HH605: Unsupported network for JSON-RPC server. Only hardhat is currently supported.
```

But if i change the script to use hardhat instead of localhost:
```json
    "deploy": "hardhat ignition deploy ignition/modules/marketplace.js --network hardhat",
    "predeploy": "hardhat run scripts/mine.js --network hardhat",
```
I get 
```bash
1] Futures failed during execution:
[1]  - Marketplace#Marketplace: Simulating the transaction failed with error: Reverted with custom error Proofs_InsufficientBlockHeight()
[1]  - Marketplace#TestMarketplace: Simulating the transaction failed with error: Reverted with custom error Proofs_InsufficientBlockHeight()
[1] 
```
Which seems like the mine.js script ran against a different network?? So weird

emizzle:
> It will recreate the deployment files for the contract address.

I'm wondering if we should run this migration script and commit the artifacts in the `ignition/deployments` folder as part of this PR.

package.json
emizzle: Anyway, you already mentioned this in a previous comment. I just thought it was a bit odd, as I'd imagined these network names should match 🤷‍♂️

package.json
emizzle: Ok, so I tried out other options and none of them work as well as the `concurrently` option. I don't love the `[1]` and [2]` output, so maybe we could add `--prefix none`?

ignition/modules/token.js
2-towns: Yes I think we can remove it but then, the contract address will change: 

[1] Marketplace#Marketplace - 0xCf7Ed3AccA5a467e9e704C703E8D87F634fB0Fc9
[1] Marketplace#TestMarketplace - 0xDc64a140Aa3E981100a9becA4E685f962f0cF6C9

instead of 

[1] Marketplace#Marketplace - 0x322813Fd9A801c5507c9de605d63CEA4f2CE6c44
[1] Marketplace#TestMarketplace - 0xa85233C63b9Ee964Add6F2cffe00Fd84eb32338f

package.json
2-towns: > Which seems like the mine.js script ran against a different network?? So weird 

Yes if you switch to `--network hardhat`, it will not connect to your `hardhat node`, but run the deployment in an internal memory instance, the `dry-run` I mentioned previously.

package.json
2-towns: Yep, I added the timestamp, could be useful as well, wdyt ?

ignition/modules/token.js
emizzle: Wait, this script runs after the contract is already deployed, so how would it affect the contract deployment address?

markspanbroek (approved):
Great work @2-towns! 

Can you also remove the `.tool-versions` file that still points to an old version of nodejs? (It is used by the asdf version manager)

markspanbroek:
> do we want to keep the previous ABI contract data or are we okay to keep the new generated one ?

I would be ok with deleting the old deployment info, and having a readme in the new ignition/deployments folder that points the reader to the last commit where the old ABI contract data is still available.

ignition/modules/token.js
emizzle: > The script mints TST tokens, not ETH, for use in tests.

Yes, good point. Perhaps we should rename the `id` to ``SendingTestTokens_${i}`` instead?

2-towns:
@markspanbroek I rebased the master which includes now the Vault and updated the Vault tests. Can you review this part please ?

test/Vault.tests.js
```diff
@@ -354,8 +419,10 @@ describe("Vault", function () {
       }
 
       it("moves tokens over time", async function () {
+        await setAutomine(true)
```
markspanbroek, line 422: I would not use auto mining here; these tests are very sensitive to timing so I used manual mining. Manual mining ensures that the block time does not increase between two transactions, which can happen every now and then if the system clock happens to tick to the next second between the two transactions.

test/evm.js
```diff
@@ -1,59 +1,57 @@
...
+    await snapshot.restore()
+    await setNextBlockTimestamp(time)
+    await provider.send("evm_setAutomine", [automine])
+  }
 }
 
-/**
- * Enables or disables Hardhat's automine mode.
- *
- * When automine mode is disabled, transactions that revert are silently ignored!
```
markspanbroek, line 22: I'd like to keep this warning around, unless the new hardhat no longer has this behaviour?

markspanbroek (commented):
The vault changes look good, except for the automine changes which I expect to make the tests unstable.

ignition/modules/token.js
veaceslavdoina: @2-towns, should we do the same with files from https://github.com/codex-storage/codex-contracts-eth/pull/226 (needs to be merged)?

2-towns: > @2-towns, should we do the same with files from #226 (needs to be merged)?

Do you mean, reference the commit hash for the Linea artifacts already deployed ?

veaceslavdoina: > > @2-towns, should we do the same with files from #226 (needs to be merged)?
> 
> Do you mean, reference the commit hash for the Linea artifacts already deployed ?

As I see, files from deployments were deleted - https://github.com/codex-storage/codex-contracts-eth/pull/231/files#diff-51b4fca6d528a9a916aff22a4d8f4854652fcdf739e226255d4f6ef241f89093

Should we merge a PR and delete artifacts as well?

2-towns: > > > @2-towns, should we do the same with files from #226 (needs to be merged)?
> > 
> > 
> > Do you mean, reference the commit hash for the Linea artifacts already deployed ?
> 
> As I see, files from deployments were deleted - https://github.com/codex-storage/codex-contracts-eth/pull/231/files#diff-51b4fca6d528a9a916aff22a4d8f4854652fcdf739e226255d4f6ef241f89093
> 
> Should we merge a PR and delete artifacts as well?

Yes I guess we should first merge the Linea PR and then this one.

test/Vault.tests.js
markspanbroek: O, that's strange. I'll take a look later today.

markspanbroek: I figured it out; hardhat now has a weird optimization whereby they set the gas limit of transactions to the gas limit of the block. This allows only a single transaction in a block, which breaks these tests. You can change this by setting `gas: "auto"` in the config: https://hardhat.org/hardhat-network/docs/explanation/mining-modes#mining-multiple-transactions-in-one-block

2-towns: > I figured it out; hardhat now has a weird optimization whereby they set the gas limit of transactions to the gas limit of the block. This allows only a single transaction in a block, which breaks these tests. You can change this by setting `gas: "auto"` in the config: https://hardhat.org/hardhat-network/docs/explanation/mining-modes#mining-multiple-transactions-in-one-block

Thanks Mark ! Okay I applied the gas auto configuration and reverted the auto mining changes that I introduced into the Vault tests.

test/Vault.tests.js
```diff
@@ -355,7 +420,7 @@ describe("Vault", function () {
 
       it("moves tokens over time", async function () {
         await vault.flow(fund, account1, account2, 2)
-        mine()
+        await mine();
```
markspanbroek, line 423: I think you accidentally added semi-colons to all the `mine()` statements.

markspanbroek (approved):
Yes, vault changes look good now, thanks @2-towns!

package.json
```diff
@@ -4,29 +4,31 @@
...
-    "start": "hardhat node --export deployment-localhost.json",
+    "start": "concurrently --names \"hardhat,deployment\"  --prefix \"[{time} {name}]\" \"hardhat node\" \"sleep 2 &&  npm run deploy\"",
     "compile": "hardhat compile",
-    "format": "prettier --write contracts/*.sol contracts/**/*.sol test/**/*.js",
-    "format:check": "prettier --check contracts/*.sol contracts/**/*.sol test/**/*.js",
+    "format": "prettier --write test/**/*.js --plugin=prettier-plugin-solidity contracts/**/*.sol ",
+    "format:check": "prettier --check test/**/*.js --plugin=prettier-plugin-solidity contracts/**/*.sol",
     "lint": "solhint contracts/**.sol",
-    "deploy": "hardhat deploy",
+    "deploy": "hardhat ignition deploy ignition/modules/marketplace.js --network localhost",
```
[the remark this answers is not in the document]
2-towns: If you run `hardhat ignition deploy` without the `--network`, it will run a memory instance. 

So, do you want a command for each environment ? `deploy-localhost`, `deploy-testnet`....

Otherwise, to deploy on a new environment you can just use `npx` command.

.gitignore
```diff
@@ -4,3 +4,6 @@ artifacts
 deployment-localhost.json
 crytic-export
 .certora_internal
+coverage
+coverage.json
+ignition/deployments
```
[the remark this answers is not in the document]
2-towns: Ok I added the deployments  folder.  

I think we can ignore `journal.jsonl` we may change too frequently and the `build-info` which is big (17 Mo).

package.json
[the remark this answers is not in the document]
AuHau: How does it work with the local hardhat network? When you do `hardhat start` does it also deploy the contracts? In what cases you actually need to run the localhost deploy command?

AuHau:
> I am not sure.

Well, I would prefer to have this figured out and verified before merging this PR. I was already solving this exact problem not long ago (because of the Solidity update) and we will need to solve it once we release, and it's better to be ready now before on release day.

2-towns:
> Well, I would prefer to have this figured out and verified before merging this PR. I was already solving this exact problem not long ago (because of the Solidity update) and we will need to solve it once we release, and it's better to be ready now before on release day.

Okay so should I deploy the Marketplace on Testnet to see what's behavior ?

AuHau:
Yeah, that would be good idea. But I can already tell you know that it will re-deploy the Test Token ;-) So while at it already please have a look on how to point it to the current address that we use.

2-towns:
> Yeah, that would be good idea. But I can already tell you know that it will re-deploy the Test Token ;-) So while at it already please have a look on how to point it to the current address that we use.

Okay I will try to deploy it. 

We can re-use a contract using [contractAt](https://hardhat.org/ignition/docs/guides/creating-modules#using-an-existing-contract).

Readme.md
```diff
@@ -25,6 +25,60 @@ To start a local Ethereum node with the contracts deployed, execute:
 This will create a `deployment-localhost.json` file containing the addresses of
 the deployed contracts.
 
+To deploy the marketplace you can define `HARDHAT_NETWORK` and run the following command:
+
+```bash
+npm run deploy
```
AuHau, line 31:
```suggestion
npm run deploy -- --network localhost
```

This is btw. a way how you can pass other parameters to the underlying `deploy` script, so you can remove the "Or you can use npx directly" cases. That is btw. why I have asked you to remove originally the `--network localhost` from the `deploy` script because you can then specify it yourself.

ignition/modules/token.js
```diff
@@ -0,0 +1,32 @@
...
+const MAX_ACCOUNTS = 20
+const MINTED_TOKENS = 1_000_000_000_000_000n
+
+module.exports = buildModule("Token", (m) => {
+  const deployer = m.getAccount(0)
+
+  let token
+
+  if (process.env.TOKEN_ADDRESS) {
+    token = m.contractAt("TestToken", process.env.TOKEN_ADDRESS, {})
```
AuHau, line 12:
```suggestion
    console.log("Using existing TestToken on address: ", process.env.TOKEN_ADDRESS)
    token = m.contractAt("TestToken", process.env.TOKEN_ADDRESS, {})
```

Let's add some logging so the deployer knows that this contract will be using the existing token. Just in case somebody, for example, forgets to clear the env. variable and can be then unpleasantly surprised.

package.json
```diff
@@ -4,29 +4,32 @@
   "scripts": {
     "test": "npm run lint && hardhat test",
     "fuzz": "hardhat compile && fuzzing/fuzz.sh",
-    "start": "hardhat node --export deployment-localhost.json",
+    "start": "concurrently --names \"hardhat,deployment\"  --prefix \"[{time} {name}]\" \"hardhat node\" \"sleep 2 && npm run mine && HARDHAT_NETWORK=localhost npm run deploy\"",
```
AuHau, line 7: wooow really?! Hardhat will not deploy the contracts automatically when starting the local network? That is weird. I mean, now it works like that... Well you managed to work around it which is great 👍

AuHau (approved):
Great! Thanks for testing the deployment and doing the tweaks.

I left some comments, but nothing blocking ;-)

package.json
```diff
@@ -4,29 +4,32 @@
...
     "compile": "hardhat compile",
-    "format": "prettier --write contracts/*.sol contracts/**/*.sol test/**/*.js",
-    "format:check": "prettier --check contracts/*.sol contracts/**/*.sol test/**/*.js",
+    "format": "prettier --write test/**/*.js --plugin=prettier-plugin-solidity contracts/**/*.sol ",
+    "format:check": "prettier --check test/**/*.js --plugin=prettier-plugin-solidity contracts/**/*.sol",
     "lint": "solhint contracts/**.sol",
-    "deploy": "hardhat deploy",
+    "deploy": "hardhat ignition deploy ignition/modules/marketplace.js --network $HARDHAT_NETWORK",
+    "deploy:reset": "hardhat ignition deploy ignition/modules/marketplace.js --network $HARDHAT_NETWORK --reset",
```
AuHau, line 13: Sooo I am a bit against this. I originally thought that `HARDHAT_NETWORK` is an env. variable that supports Hardhat directly, and not one that you added support for in this way. My main problem with this is multiplatform support. The npm-scripts are run, I believe, using `sh` for Mac/Linux, but for Windows, it is `cmd`, which obviously does not support `$`. 

Actually, I think you will run into this in your `nim-codex`'s PR in the Windows CI? Obviously, if they are running in the WSL, than it would work.

So I would again come to my original suggestion that you can see here:
```suggestion
    "deploy": "hardhat ignition deploy ignition/modules/marketplace.js"
```

You can then use it as `npm run deploy -- --network XYZ --reset`, because you can [pass args to npm script](https://stackoverflow.com/questions/11580961/sending-command-line-arguments-to-npm-script).

To address @veaceslavdoina point about env. variables - how do you use them now? I was looking into Dockerfiles and I am not sure how do you actually configure the network parameter for the `deploy.Dockerfile`? I would suggest adding this sort of support for the network env. variable in the `deploy.sh` script.

AuHau (changes requested):

package.json
2-towns: Okay I updated the package.json and the documentation.

AuHau (approved):
Thanks, now LGTM 🙏

hardhat.config.js
```diff
@@ -24,7 +23,11 @@ module.exports = {
   networks: {
     hardhat: {
       tags: ["local"],
-      allowBlocksWithSameTimestamp: true
+      allowBlocksWithSameTimestamp: true,
+      gas: "auto",
+    },
+    localhost: {
```
AuHau, line 29: Why did you add this network?

ignition/modules/endian.js
```diff
@@ -0,0 +1,15 @@
+const { buildModule } = require("@nomicfoundation/hardhat-ignition/modules")
+
+module.exports = buildModule("Endian", (m) => {
+  const deployer = m.getAccount(0)
```
AuHau, line 4: Btw. isn't this by default? IMHO we don't need it?

AuHau:
While working on the Upgradability, I ran into nonce issues with Hardhat Ignition. I originally thought there is some problem with my changes, but now I have run few times tests on this branch as well, and I am getting sporadically the same errors. @2-towns have you seen this before?

```
  1) Marketplace
       "before each" hook for "emits event when storage is requested":
     IgnitionError: IGN403: You have sent transactions from 0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266 and they interfere with Hardhat Ignition. Please wait until they get 1 confirmations before running Hardhat Ignition again.
      at getNonceSyncMessages (node_modules/@nomicfoundation/ignition-core/src/internal/execution/nonce-management/get-nonce-sync-messages.ts:118:15)
      at ExecutionEngine._syncNonces (node_modules/@nomicfoundation/ignition-core/src/internal/execution/execution-engine.ts:253:31)
      at ExecutionEngine.executeModule (node_modules/@nomicfoundation/ignition-core/src/internal/execution/execution-engine.ts:77:23)
      at Deployer.deploy (node_modules/@nomicfoundation/ignition-core/src/internal/deployer.ts:196:25)
      at Proxy.deploy (node_modules/@nomicfoundation/hardhat-ignition-ethers/src/ethers-ignition-helper.ts:154:20)
      at Context.<anonymous> (test/Marketplace.test.js:142:48)



IgnitionError: IGN405: The next nonce for 0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266 should be 20, but is 4. Please make sure not to send transactions from 0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266 while running this deployment and try again.
    at JsonRpcNonceManager.getNextNonce (/Users/adam/Projects/status/codex-contracts-eth/node_modules/@nomicfoundation/ignition-core/src/internal/execution/nonce-management/json-rpc-nonce-manager.ts:51:13)
    at sendTransactionForOnchainInteraction (/Users/adam/Projects/status/codex-contracts-eth/node_modules/@nomicfoundation/ignition-core/src/internal/execution/future-processor/helpers/network-interaction-execution.ts:124:34)
    at sendTransaction (/Users/adam/Projects/status/codex-contracts-eth/node_modules/@nomicfoundation/ignition-core/src/internal/execution/future-processor/handlers/send-transaction.ts:87:18)
    at FutureProcessor.processFuture (/Users/adam/Projects/status/codex-contracts-eth/node_modules/@nomicfoundation/ignition-core/src/internal/execution/future-processor/future-processor.ts:117:9)
    at ExecutionEngine._executeBatch (/Users/adam/Projects/status/codex-contracts-eth/node_modules/@nomicfoundation/ignition-core/src/internal/execution/execution-engine.ts:161:30)
    at ExecutionEngine.executeModule (/Users/adam/Projects/status/codex-contracts-eth/node_modules/@nomicfoundation/ignition-core/src/internal/execution/execution-engine.ts:122:25)
    at Deployer.deploy (/Users/adam/Projects/status/codex-contracts-eth/node_modules/@nomicfoundation/ignition-core/src/internal/deployer.ts:196:25)
    at Proxy.deploy (/Users/adam/Projects/status/codex-contracts-eth/node_modules/@nomicfoundation/hardhat-ignition-ethers/src/ethers-ignition-helper.ts:154:20)
```

2-towns:
> While working on the Upgradability, I ran into nonce issues with Hardhat Ignition. I originally thought there is some problem with my changes, but now I have run few times tests on this branch as well, and I am getting sporadically the same errors. @2-towns have you seen this before?
> 
> ```
>   1) Marketplace
>        "before each" hook for "emits event when storage is requested":
>      IgnitionError: IGN403: You have sent transactions from 0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266 and they interfere with Hardhat Ignition. Please wait until they get 1 confirmations before running Hardhat Ignition again.
>       at getNonceSyncMessages (node_modules/@nomicfoundation/ignition-core/src/internal/execution/nonce-management/get-nonce-sync-messages.ts:118:15)
>       at ExecutionEngine._syncNonces (node_modules/@nomicfoundation/ignition-core/src/internal/execution/execution-engine.ts:253:31)
>       at ExecutionEngine.executeModule (node_modules/@nomicfoundation/ignition-core/src/internal/execution/execution-engine.ts:77:23)
>       at Deployer.deploy (node_modules/@nomicfoundation/ignition-core/src/internal/deployer.ts:196:25)
>       at Proxy.deploy (node_modules/@nomicfoundation/hardhat-ignition-ethers/src/ethers-ignition-helper.ts:154:20)
>       at Context.<anonymous> (test/Marketplace.test.js:142:48)
> 
> 
> 
> IgnitionError: IGN405: The next nonce for 0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266 should be 20, but is 4. Please make sure not to send transactions from 0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266 while running this deployment and try again.
>     at JsonRpcNonceManager.getNextNonce (/Users/adam/Projects/status/codex-contracts-eth/node_modules/@nomicfoundation/ignition-core/src/internal/execution/nonce-management/json-rpc-nonce-manager.ts:51:13)
>     at sendTransactionForOnchainInteraction (/Users/adam/Projects/status/codex-contracts-eth/node_modules/@nomicfoundation/ignition-core/src/internal/execution/future-processor/helpers/network-interaction-execution.ts:124:34)
>     at sendTransaction (/Users/adam/Projects/status/codex-contracts-eth/node_modules/@nomicfoundation/ignition-core/src/internal/execution/future-processor/handlers/send-transaction.ts:87:18)
>     at FutureProcessor.processFuture (/Users/adam/Projects/status/codex-contracts-eth/node_modules/@nomicfoundation/ignition-core/src/internal/execution/future-processor/future-processor.ts:117:9)
>     at ExecutionEngine._executeBatch (/Users/adam/Projects/status/codex-contracts-eth/node_modules/@nomicfoundation/ignition-core/src/internal/execution/execution-engine.ts:161:30)
>     at ExecutionEngine.executeModule (/Users/adam/Projects/status/codex-contracts-eth/node_modules/@nomicfoundation/ignition-core/src/internal/execution/execution-engine.ts:122:25)
>     at Deployer.deploy (/Users/adam/Projects/status/codex-contracts-eth/node_modules/@nomicfoundation/ignition-core/src/internal/deployer.ts:196:25)
>     at Proxy.deploy (/Users/adam/Projects/status/codex-contracts-eth/node_modules/@nomicfoundation/hardhat-ignition-ethers/src/ethers-ignition-helper.ts:154:20)
> ```

Nope, first time I see those errors.

ignition/modules/endian.js
2-towns: I am not sure to understand. The `Endian` module is used by the Marketplace right ?  

So this ignition module is used in `Endian.test.js`.

hardhat.config.js
AuHau: Oh right, I guess `hardhat` is then used for testing?

ignition/modules/endian.js
2-towns: Oh sorry. 

Yes you are right, this is the default, so I removed it in my last push.

PR #234, opened 2025-04-15
fix(slot-reservations): Allows slot to be reserved when in repair

emizzle:
Previous to when SlotState.Repair was implemented, slots in repair would be considered free and the slots could be reserved in this state. Now that SlotState.Repair has been implemented, the `canReserveSlot` needs to check that the SlotState is in Repair or is Free before allowing reservation.


markspanbroek (approved):
Nice, good catch @emizzle!

PR #235, opened 2025-04-17
fix(slot reservations): clear AddressSet instead of delete

emizzle:
Deleting an AddressSet causes corrupted memory. Each address must be removed individually, which is OK to do since there is a maxReservations parameter that keeps this number small.

https://docs.openzeppelin.com/contracts/5.x/api/utils#EnumerableSet

### DIFF
--- a/contracts/Marketplace.sol
+++ b/contracts/Marketplace.sol
@@ -376,7 +376,7 @@ contract Marketplace is SlotReservations, Proofs, StateRetrieval, Endian {
     context.fundsToReturnToClient += _slotPayout(requestId, slot.filledAt);
 
     _removeFromMySlots(slot.host, slotId);
-    _clear(_reservations[slotId]); // We purge all the reservations for the slot
+    _reservations[slotId].clear(); // We purge all the reservations for the slot
     slot.state = SlotState.Repair;
     slot.filledAt = 0;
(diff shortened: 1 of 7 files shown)


markspanbroek (approved):
Good catch, thanks @emizzle!

AuHau (approved):
LGTM, great catch!

2-towns:
For some reason, it doesn't work when I tried to write the functional tests for slot repair. OpenZeppelin has provided an update recently and [EnumerableSet now has a clear function](https://docs.openzeppelin.com/contracts/5.x/api/utils#EnumerableSet-clear-struct-EnumerableSet-AddressSet-). So my suggestion is to update to the latest version (i.e., 5.0.2 to 5.0.3) and use this function.

emizzle:
> For some reason, it doesn't work when I tried to write the functional tests for slot repair. OpenZeppelin has provided an update recently and [EnumerableSet now has a clear function](https://docs.openzeppelin.com/contracts/5.x/api/utils#EnumerableSet-clear-struct-EnumerableSet-AddressSet-). So my suggestion is to update to the latest version (i.e., 5.0.2 to 5.0.3) and use this function.

Well done @2-towns, thanks for the find and commit 🙌

PR #238, opened 2025-05-14
fix(vault): do no allow reuse of fund ids

markspanbroek:
Fixes an attack where all tokens can be drained from the Vault by allowing a token flow to persist after a Fund is deleted.


emizzle (approved):
Nice work everyone!

PR #249, opened 2025-06-13
fix: use trySafeTransferFrom instead of transferFrom

2-towns:
When using `transferFrom`, the custom error `ERC20InsufficientAllowance` may be raised. This is potentially problematic, because I didn't find this error declared like other custom errors in the Marketplace, meaning that the error would not be recognized if it is raised (if I understand correctly).

Additionally, [we are checking that the transferFrom returns false](https://github.com/codex-storage/codex-contracts-eth/blob/08e91c2443693d2f083adaa9a70baedbbbcba10a/contracts/Marketplace.sol#L690-L691) in order to determine if the operation succeeded or not, [as it is defined in the spec](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/5c4b75a9cdda99af901b8c153d0d1e8a90b0139a/contracts/token/ERC20/IERC20.sol#L54-L67) but the [implementation is different](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/5c4b75a9cdda99af901b8c153d0d1e8a90b0139a/contracts/token/ERC20/ERC20.sol#L158-L167) and `false` is never returned as mentioned in [this issue](https://github.com/OpenZeppelin/openzeppelin-contracts/issues/4777#issuecomment-1869719904).

This PR replaces `transferFrom` by `trySafeTransferFrom` to handle the potential errors and returns `Marketplace_TransferFailed` if something went wrong.

### DIFF
--- a/contracts/Marketplace.sol
+++ b/contracts/Marketplace.sol
@@ -2,6 +2,7 @@
 pragma solidity 0.8.28;
 
 import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
 import "@openzeppelin/contracts/utils/math/Math.sol";
 import "@openzeppelin/contracts/utils/structs/EnumerableSet.sol";
 import "./Configuration.sol";
@@ -45,6 +46,7 @@ contract Marketplace is SlotReservations, Proofs, StateRetrieval, Endian {
   using EnumerableSet for EnumerableSet.AddressSet;
   using Requests for Request;
   using AskHelpers for Ask;
+  using SafeERC20 for IERC20;
 
   IERC20 private immutable _token;
   MarketplaceConfig private _config;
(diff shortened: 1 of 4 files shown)


markspanbroek:
I did something similar in the Vault PR: https://github.com/codex-storage/codex-contracts-eth/pull/223/commits/341b303789a895fff13eb362b41fa9a226ec0cbc

There I used `safeTransferFrom()`, which can indeed raise an ERC20 error. I kind of like that, because it gives more detailed information about what went wrong (infufficient balance? insufficient allowance? transfer returned false?) As you mentioned we could declare these ERC20 errors as custom errors to allow for better error logs.

2-towns:
> I did something similar in the Vault PR: [341b303](https://github.com/codex-storage/codex-contracts-eth/commit/341b303789a895fff13eb362b41fa9a226ec0cbc)
> 
> There I used `safeTransferFrom()`, which can indeed raise an ERC20 error. I kind of like that, because it gives more detailed information about what went wrong (infufficient balance? insufficient allowance? transfer returned false?) As you mentioned we could declare these ERC20 errors as custom errors to allow for better error logs.

Okay so I can add the ERC20 errors in nim-codex and use `safeTransferFrom` instead. But then, do you think we should keep this PR to be merged quickly or should we close it in favor of the vault PR ?

markspanbroek:
> Okay so I can add the ERC20 errors in nim-codex and use `safeTransferFrom` instead. But then, do you think we should keep this PR to be merged quickly or should we close it in favor of the vault PR ?

I'll leave that up to you, I don't think there's a rush to get this fixed so we can wait for the Vault PR, but I also don't mind merging this PR and then removing the transfer change from vault PR.

